PR #17, opened 2018-07-16
Bugfix/terminate users

meletios:


member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -277,6 +285,12 @@ public void on(MemberCancellationEvent e) {
         this.status = MemberStatus.TERMINATED;
     }
 
+
+    @EventSourcingHandler
+    public void on(InsuranceCancellationEvent e) {
+        log.info("Cancel insurance with id {} for member {}", e.getInsuranceId(), e.getMemberId());
```
hedvigoscar: Should we note this event on the aggregate, or are we happy with noting it for future reference?

member-service/src/main/java/com/hedvig/memberservice/commands/MemberCancelInsuranceCommand.java
```diff
@@ -10,5 +10,6 @@
     @TargetAggregateIdentifier
     Long memberId;
 
+
```
hedvigoscar: Was this accidental? :)

PR #19, opened 2018-08-02
git hash build id!

hedvigoscar:
also sneaking in some other stuff for good measure


hedvigoscar:
@johantjellden I'm merging this now!

PR #27, opened 2018-08-21
Add memberId to sentry user and tags, improve stacktrace

johantjellden:


member-service/src/main/resources/sentry.properties
```diff
@@ -0,0 +1,2 @@
+mdctags=foo,bar
```
meletios: change dat please

PR #46, opened 2018-10-12
Feature/single sign session per member

johantjellden:
Keep only one signsession active per member .


member-service/src/main/java/com/hedvig/memberservice/enteties/BankIdSession.java
```diff
@@ -0,0 +1,54 @@
+package com.hedvig.memberservice.enteties;
```
meletios: We should rename 'enteties' to 'entities'

PR #133, opened 2018-12-18
Remove apartmentNumber from members address

johantjellden:


member-service/src/main/java/com/hedvig/memberservice/web/AddressLookup.kt
```diff
@@ -42,8 +42,7 @@ class AddressLookup(private val client: BisnodeClient) {
             
             val addr = if (person.addressList.size > 0 && person.addressList[0].streetName != null) {
                 val address = person.addressList[0]
```
palmfjord: mabye move this up to L42 as were using the same thing on L43

member-service/src/main/java/com/hedvig/memberservice/web/AddressLookup.kt
```diff
@@ -39,11 +39,10 @@ class AddressLookup(private val client: BisnodeClient) {
...
             val person = match.persons[0].person
-            
-            val addr = if (person.addressList.size > 0 && person.addressList[0].streetName != null) {
-                val address = person.addressList[0]
-                val apartment = if (address.apartment != null) " ${address.apartment}" else ""
-                Address("${address.streetName} ${address.streetNumber.orEmpty()}${address.entrance.orEmpty()}$apartment", address.city, address.postalCode, address.apartment)
+            val address =  person.addressList.getOrNull(0)
+            val addr = if (address?.streetName != null) {
+
+                Address("${address.streetName} ${address.streetNumber.orEmpty()}${address.entrance.orEmpty()}", address.city, address.postalCode, address.apartment)
```
palmfjord: shouldn't we use the null-safe operator here too?

johantjellden: Maybe how would you do that?

palmfjord: i mean since `address` can be null shouldn't we use `?.` here too? alternatively check so `address` isn't null before `address?.streetName != null`?

palmfjord: ☝️ a classic tankevurpa. Since `address?.streetName' will be `null` if `address` is null lol

palmfjord: nvm

PR #221, opened 2019-06-03
call end-point within product_pricing to update member first and last…

AlexSJ1990:
… name


member-service/src/main/java/com/hedvig/memberservice/query/MemberEventListener.java
```diff
@@ -126,6 +134,14 @@ void on(NameUpdatedEvent e) {
...
 
     userRepo.save(m);
+
+    EditMemberNameDto editMemberNameDto = new EditMemberNameDto(
+      String.valueOf(e.getMemberId()),
+      e.getFirstName(),
+      e.getLastName()
+    );
+
+    productApi.editMember(String.valueOf(e.getMemberId()), editMemberNameDto, "abc");
```
vonElfvin: If we know "e.getFirstName()" and "e.getLastName()" are not null then the check in productpricing may be overkill. Or we set the name to null here in the database which is not so good.

palmenhq: Also should `token` really be `abc`? looks suspicious

member-service/src/main/java/com/hedvig/memberservice/query/MemberEventListener.java
```diff
@@ -42,17 +41,20 @@
...
+  private final BotServiceImpl botServiceImpl;
 
   @Autowired
   public MemberEventListener(
     MemberRepository userRepo,
     SignedMemberRepository signedMemberRepository,
     TrackingIdRepository trackingRepo,
-    ProductApi productApi) {
+    ProductApi productApi,
+    BotServiceImpl botServiceImpl) {
```
palmenhq: no `Impl`s when we use them please :) `BotService botService` should be sufficient here

palmenhq: and then ofc we need to pull in the method we implemented in `BotServiceImpl` so its also in the `BotService` interface

palmenhq:
@vonElfvin i think for now, lets leave it. If we know we really need to backfill this then let's make sure we do that separately so we dont mess anything up hehe 😬

PR #242, opened 2019-06-24
Add a campaign profile

johantjellden:
@palmenhq I Added a feature flag to that memberservice can be deployed to prod.


member-service/src/main/java/com/hedvig/memberservice/config/AxonConfig.java
```diff
@@ -38,6 +39,7 @@
     return config;
   }
 
+  @Profile("campaign")
```
meletios: It shouldnt be only for campaigns

meletios: It should enable always

johantjellden: Fixed this.

PR #263, opened 2019-07-22
New Auth

hedvigoscar:


.editorconfig
```diff
@@ -9,3 +9,7 @@ insert_final_newline = true
 [*.java]
 indent_style = space
 indent_size = 2
+
+[*.kt]
+indent_style = space
+indent_size = 4
```
meletios: 4 or 2 🤔

hedvigoscar: Standard kotlin style is 4 afaik 🙂

PR #319, opened 2019-10-13
return errorResponseDto if ssn from rapio is a signed member already

AlexSJ1990:


member-service/src/main/java/com/hedvig/memberservice/web/v2/SignController.java
```diff
@@ -33,11 +34,16 @@
...
-     val result = signingService.signUnderwriterQuote(memberId, underwriterQuoteSignRequest);
-    return ResponseEntity.ok(new UnderwriterQuoteSignResponse(result.getSignId(),result.getMemberIsSigned()));
+  public ResponseEntity<?> signQuotesFromUnderwriter(
+    @RequestHeader("hedvig.token") final long memberId,
+    @RequestBody UnderwriterQuoteSignRequest underwriterQuoteSignRequest) {
+
+    try {
+      val result = signingService.signUnderwriterQuote(memberId, underwriterQuoteSignRequest);
+      return ResponseEntity.ok(new UnderwriterQuoteSignResponse(result.getSignId(),result.getMemberIsSigned()));
+    } catch(Exception exception) {
```
palmenhq: can we be a bit more granular in the exception we throw here? especially considering we're assuming member already exists if we catch it

member-service/src/main/java/com/hedvig/memberservice/web/v2/SignController.java
```diff
@@ -33,11 +34,16 @@
...
+      return ResponseEntity.status(402).body(new ErrorResponseDto("Member already has an existing product"));
```
palmenhq: I _really_ don't like using 402. This should be either 400 (bad request) possibly 422 (unprocessable entity) imo. Referring to [IETF RFC 7231](https://tools.ietf.org/html/rfc7231#section-6.5.2) and [IETF RFC 4918](https://tools.ietf.org/html/rfc4918#section-11.2) - the official HTTP standards afaik:

> **402**
> The 402 (Payment Required) status code is reserved for future use.

> **422**
>  The 422 (Unprocessable Entity) status code means the server
   understands the content type [...] **but was unable to process the contained
   instructions**.

PR #321, opened 2019-10-19
attribution header in hello hedvig

FredrikAreschoug:


member-service/src/main/java/com/hedvig/memberservice/web/MembersController.java
```diff
@@ -117,6 +108,10 @@ public MembersController(MemberRepository repo, CommandGateway commandGateway,
             return memberId;
         });
 
+        if (attributionCode != null && !attributionCode.isEmpty()) {
+            commandGateway.send(new AssignAttributionCodeCommand(id, attributionCode));
```
palmenhq: I think there's a high chance we get a race condition between what finishes first - `commandGateway.send(new CreateMemberCommand(memberId))` (which isn't even `sendAndWait`) and this line, which makes it likely the [`MemberListener`](https://github.com/HedvigInsurance/member-registry-service/pull/321/files#diff-76a591be00e12d48b3f326fda4966634R202) might not find the member at first which'll throw an exception, since the `@EventHandler`s aren't synchronous. Could we not upcast the `CreateMemberCommand` and amend the attribution code rather, or?

johantjellden: We don't need to upcast commands, since they are not persisted we can just rename them or add values to them as we want.

palmenhq: Ah yeah, of course! update* is what i ment

FredrikAreschoug: True! that was my first approach but didn't get it to work for some reason will look in to it again

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -66,6 +66,13 @@ public MemberAggregate(
   @CommandHandler
   public MemberAggregate(CreateMemberCommand command) {
     apply(new MemberCreatedEvent(command.getMemberId(), MemberStatus.INITIATED));
+
+    if (command.getAttributionCode() != null) {
+      log.info("Assign attributionCode for member {}, new attributionCode: {}", command.getMemberId(),
+              command.getAttributionCode());
+
+      apply(new AssignAttributionCodeEvent(command.getMemberId(), command.getAttributionCode()));
```
palmenhq: awesome! However i think we should assign the first `apply` to a variable and call `.andThenApply(...)` on it - so we don't have to deal with race conditions

member-service/src/main/java/com/hedvig/memberservice/web/MembersController.java
```diff
@@ -101,7 +92,7 @@ public MembersController(MemberRepository repo, CommandGateway commandGateway,
     }
 
     @PostMapping("/helloHedvig")
-    public ResponseEntity<String> helloHedvig() throws Exception {
+    public ResponseEntity<String> helloHedvig(@RequestHeader(required = false) String attributionCode) throws Exception {
```
palmenhq: Hm also another though, what'll be the name of this header? i think something like `x-attribution-code` or so would make more sense than `attributionCode` (which i'm kind of assuming itll resolve to)

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
FredrikAreschoug: Didn't know that existed, awesome

PR #322, opened 2019-10-21
Cleanup non-signed users on customer.io with the same SSN or email as signed users

vonElfvin:


member-service/src/main/java/com/hedvig/integration/customerIO/CustomerIOEventListener.kt
```diff
@@ -0,0 +1,43 @@
...
+    fun on(event: MemberSignedEvent) {
+        val member = memberRepository.findById(event.id).get()
+        val membersToDeleteFromCustomerIO = getNonSignedMembersWithSameSsnOrEmail(
+            memberId = member.id,
+            ssn = member.ssn,
+            email = member.email
+        )
+        membersToDeleteFromCustomerIO.forEach { memberToRemove ->
+            try {
+                customerIO.deleteUser(userId = memberToRemove.id.toString())
```
palmenhq: just for the record CIO has a rate limit of 30 requests per second so i think we should throttle this so we don't lock ourselves out

member-service/src/main/java/com/hedvig/integration/customerIO/CustomerIOEventListener.kt
```diff
@@ -0,0 +1,43 @@
...
+            } catch (exception: Exception) {
+                logger.error { "Failed to delete member=${memberToRemove.id} from customer.io (exception=$exception)" }
+            }
+        }
+    }
+
+    private fun getNonSignedMembersWithSameSsnOrEmail(memberId: Long, ssn: String, email: String): List<MemberEntity> =
+        memberRepository.findBySsnOrEmail(ssn, email)
+            .filter { member -> member.id != memberId } // To avoid race condition
+            .filter { member -> member.status != MemberStatus.SIGNED }
```
palmenhq: any particular reason for not using sql for this?

member-service/src/main/java/com/hedvig/integration/customerIO/CustomerIOEventListener.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.hedvig.memberservice.query.MemberRepository
+import mu.KotlinLogging
+import org.axonframework.config.ProcessingGroup
+import org.axonframework.eventhandling.EventHandler
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.context.annotation.Profile
+
+@Profile("customer.io")
+@ProcessingGroup("CustomerIO")
+class CustomerIOEventListener @Autowired constructor(
```
palmenhq: i think actually we don't need the `@Autowired constructor` here  since spring _should_ figure out it only has one constructor and inject the dependencies through that?

member-service/src/main/java/com/hedvig/integration/customerIO/CustomerIOEventListener.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.hedvig.memberservice.query.MemberEntity
+import com.hedvig.memberservice.query.MemberRepository
+import mu.KotlinLogging
+import org.axonframework.config.ProcessingGroup
+import org.axonframework.eventhandling.EventHandler
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.context.annotation.Profile
+
+@Profile("customer.io")
+@ProcessingGroup("CustomerIO")
```
palmenhq: I believe we might need a `@Component` annotation also for spring to find this guy

member-service/src/main/java/com/hedvig/memberservice/services/segmentpublisher/EventListener.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.axonframework.eventhandling.EventHandler
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.context.annotation.Profile
+import org.springframework.stereotype.Component
+import java.util.Objects
+
+@Component
+@Profile("customer.io")
+@ProcessingGroup("SegmentProcessorGroup")
+class EventListener @Autowired constructor(
```
palmenhq: Same here, i _think_ `@Autowired constructor` is redundant

member-service/src/main/java/com/hedvig/integration/customerIO/CustomerIOEventListener.kt
vonElfvin: added a thread.sleep now

meletios:
Totally agree with Mr palmen here, its better/safer if we mark them as duplicates instead of delete them.

vonElfvin:
Forgot to comment before merging, after consultation with IEX and Growth they preferred the deletion strategy. Where the risk of deleting someone that should not be deleted is extremely low. I'll see how it behaves in staging and evaluate.

PR #328, opened 2019-10-29
Store accept language on member entity

FredrikAreschoug:


member-service/src/main/java/com/hedvig/memberservice/commands/CreateMemberCommand.kt
```diff
@@ -1,5 +1,6 @@
 package com.hedvig.memberservice.commands
 
 data class CreateMemberCommand(
-  val memberId: Long
+    val memberId: Long,
```
meletios:
```suggestion
    @TargetAggregateIdentifier
    val memberId: Long,
```

member-service/src/main/java/com/hedvig/memberservice/web/AppleController.java
```diff
@@ -60,7 +60,7 @@ public AppleController(MemberRepository memberRepository, CommandGateway command
       return ResponseEntity.badRequest().build();
     }
 
-    commandGateway.sendAndWait(new CreateMemberCommand(appleMemberId));
+    commandGateway.sendAndWait(new CreateMemberCommand(appleMemberId, null));
```
meletios: Do you need to pass a `null` here since we already initiated as null in the class?

member-service/src/main/java/com/hedvig/memberservice/web/MembersController.java
```diff
@@ -203,4 +194,12 @@ public MembersController(MemberRepository repo, CommandGateway commandGateway,
 
         return ResponseEntity.accepted().build();
     }
+
+    @PostMapping("/language")
```
meletios:
```suggestion
    @PostMapping("/language/update")
```

member-service/src/main/java/com/hedvig/memberservice/web/AppleController.java
FredrikAreschoug: Yes, java dosen't care for those fancy kotlin features :)

PR #332, opened 2019-11-05
Add endpoint to backfill all UC debts

vonElfvin:


member-service/src/main/java/com/hedvig/personservice/debts/DebtService.kt
```diff
@@ -22,6 +25,16 @@ class DebtService @Autowired constructor(
         }
     }
 
+    fun checkAllPersonDebts() {
+        val allSignedSsn = memberRepository.findByStatus(MemberStatus.SIGNED)
+            .filter { member -> member.ssn != null }
```
palmenhq: `.mapNotNull`?

palmenhq: Could we do this filtering with the sql query instead? This seems expensive

vonElfvin: I changed to mapNoNull. Since it's 0 SIGNED members in prod and about 16 in staging with ssn = null I don't think this is expensive. The reason I'm using mapNotNull is for it to not crash in when testing it.

member-service/src/main/java/com/hedvig/personservice/debts/DebtService.kt
```diff
@@ -22,6 +25,15 @@ class DebtService @Autowired constructor(
         }
     }
 
+    fun checkAllPersonDebts() {
```
AlexSJ1990: i would call this checkDebtForAllPersons

PR #336, opened 2019-11-25
Fraud flag on person

vonElfvin:
Include a green, amber or red flag of person based on fraudulent status of member with same person

Dependency of https://github.com/HedvigInsurance/account-service/pull/51

### DIFF
--- a/member-service/src/main/java/com/hedvig/personservice/persons/PersonService.kt
+++ b/member-service/src/main/java/com/hedvig/personservice/persons/PersonService.kt
@@ -88,7 +88,7 @@ class PersonService @Autowired constructor(
         }
 
         fun calculateFraudFlag(members: List<MemberEntity>): Flag {
-            return when (members.maxBy { member -> member.fraudulentStatus }!!.fraudulentStatus) {
+            return when (members.maxBy { member -> member.fraudulentStatus.severity }!!.fraudulentStatus) {
                 FraudulentStatus.CONFIRMED_FRAUD -> Flag.RED
                 FraudulentStatus.SUSPECTED_FRAUD -> Flag.AMBER
                 FraudulentStatus.NOT_FRAUD -> Flag.GREEN
--- a/member-service/src/main/java/com/hedvig/personservice/persons/PersonService.kt
+++ b/member-service/src/main/java/com/hedvig/personservice/persons/PersonService.kt
@@ -1,5 +1,7 @@
 package com.hedvig.personservice.persons
 
+import com.hedvig.memberservice.aggregates.FraudulentStatus
+import com.hedvig.memberservice.query.MemberEntity
 import com.hedvig.memberservice.query.MemberRepository
 import com.hedvig.personservice.debts.DebtService
 import com.hedvig.personservice.persons.domain.commands.RemoveWhitelistCommand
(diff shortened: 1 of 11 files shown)


member-service/src/main/java/com/hedvig/personservice/persons/PersonService.kt
```diff
@@ -61,18 +63,37 @@ class PersonService @Autowired constructor(
...
-        fun getAllFlags(person: Person): PersonFlags = PersonFlags(
-            debtFlag = DebtService.getDebtFlagByPerson(person)
+        fun calculateAllFlags(person: Person, members: List<MemberEntity>): PersonFlags = PersonFlags(
+            debtFlag = DebtService.calculateDebtFlagByPerson(person),
+            fraudFlag = calculateFraudFlag(members)
         )
 
-        fun getFlag(person: Person): Flag {
-            val personFlags = getAllFlags(person)
+        fun calculateOverallFlag(personFlags: PersonFlags): Flag {
```
AlexSJ1990: why is this only taking into consideration debtFlag?

vonElfvin: For now, we only consider a person red based on their UC, we may want to extend it in the future but after discussing with IEX only UC should considered for now.

member-service/src/main/java/com/hedvig/personservice/persons/PersonService.kt
```diff
@@ -61,18 +63,37 @@ class PersonService @Autowired constructor(
...
             return when(personFlags.debtFlag) {
                 Flag.GREEN -> Flag.GREEN
                 Flag.AMBER -> Flag.GREEN
                 Flag.RED -> Flag.RED
```
palmenhq: what? Why not just `return personflag.debtFlag`?

member-service/src/main/java/com/hedvig/personservice/persons/PersonService.kt
```diff
@@ -61,18 +63,37 @@ class PersonService @Autowired constructor(
...
             }
         }
+
+        fun calculateFraudFlag(members: List<MemberEntity>): Flag {
+            return when (members.maxBy { member -> member.fraudulentStatus }!!.fraudulentStatus) {
```
palmenhq: what is the ordering of `member.fraudulentStatus`? Feels a but shaky to depend on its order or? Maybe we could do something like

```kt
when {
  members.any { member -> member.fraudulentStatus == FraudulentStatus.CONFIRMED_FRAUD } -> Flag.RED
  members.any { member -> member.fraudulentStatus == FraudulentStatus.SUSPECTED_FRAUD } -> Flag.AMBER
...
}
```

member-service/src/main/java/com/hedvig/personservice/persons/PersonService.kt
vonElfvin: Two reasons really, in this particular calculation, on line 85, we return green flag event though the debt flag is amber. Also in the future we may want to have more logic revolving what a flag is. E.g. include the fraud flag.

palmenhq: Ah yes of course, didn't catch that haha

PR #339, opened 2019-11-29
Add datadog agent to Dockerfile

johantjellden:


johantjellden:
Will use my administrator rights ❤️

PR #340, opened 2019-12-01
Use amazoncorretto as docker image

johantjellden:
OpenJDK 11 is no longer getting supported


johantjellden:
I want to see if changing the JVM has any effect on our pods going OOM. So I will force merge this and so that I can observe any effects tomorrow.

PR #344, opened 2020-01-07
Use bank id rest api

FredrikAreschoug:
I could spend more time on updating BankIdServiceV2 and pass along more of the error codes to redis, but I rather want this out so we can test it in staging


member-service/src/main/java/com/hedvig/memberservice/web/v2/AuthControllerV2.kt
```diff
@@ -16,8 +16,8 @@ class AuthControllerV2(
         private val bankIdService: BankIdServiceV2
 ) {
     @PostMapping("auth")
-    fun auth(@RequestHeader("hedvig.token") memberId: Long): ResponseEntity<AuthResponse> {
-        val status = bankIdService.auth(memberId)
+    fun auth(@RequestHeader("hedvig.token") memberId: Long, @RequestHeader(value = "x-forwarded-for", required = false) forwardFor: String): ResponseEntity<AuthResponse> {
```
johantjellden: Nice 👍

member-service/src/main/java/com/hedvig/integration/botService/BotServiceImpl.java
```diff
@@ -28,11 +28,11 @@ public void initBotServiceSessionWebOnBoarding(Long memberId, UpdateUserContextD
   public void initBotService(Long memberId, String json) {
     logger.info("Initializing bot-service by helloHedvig request");
 
-    if (json == null) {
+    /*if (json == null) {
```
johantjellden: Should we remove this?

member-service/src/main/java/com/hedvig/memberservice/CustomClientHttpRequestInterceptor.java
```diff
@@ -34,7 +34,7 @@ private void traceRequest(HttpRequest request, byte[] body) throws IOException {
     LOGGER.trace("URI         : {}", request.getURI());
     LOGGER.trace("Method      : {}", request.getMethod());
     LOGGER.trace("Headers     : {}", request.getHeaders());
-    LOGGER.trace("Request body: {}", new String(body, "UTF-8"));
```
johantjellden: I don't really now if we should keep this interceptor.

member-service/src/main/java/com/hedvig/integration/botService/BotServiceImpl.java
FredrikAreschoug: no that's something that should not be commented out!

member-service/src/main/java/com/hedvig/memberservice/web/AuthController.kt
```diff
@@ -0,0 +1,218 @@
...
+    private val bankIdService: BankIdService) {
+
+    @PostMapping(path = ["auth"])
+    fun auth(@RequestHeader(value = "x-forwarded-for", required = false) forwardedIp: String?, @RequestBody request: BankIdAuthRequest): ResponseEntity<BankIdAuthResponse> {
+        log.info(
+            "Auth request for with memberId: ${request.memberId}", StructuredArguments.value("memberId", request.memberId))
+
+        val memberId = convertMemberId(request.memberId)
+
+        val endUserIp = if (forwardedIp?.contains(",") == true) {
```
johantjellden: Should we handle the null case on `forwardedIp`?

FredrikAreschoug: Yeah true let's respond with bad request

johantjellden: Yes but the client does not set the forwarder-for header, maybe a 500 then. Or should we  set the ip to "127.0.0.1" and log an error to sentry. So that we don't start crashing due to a misconfigured loadbalancer.

FredrikAreschoug: true let's just log it

PR #346, opened 2020-01-20
Feature/change signing to underwritter plus mandates

meletios:
- Redirects signing to underwriter instead of product-pricing
- ~~handles the creation of the member mandate~~


member-service/src/main/java/com/hedvig/memberservice/events/MemberSignedEvent.java
```diff
@@ -18,4 +15,87 @@
...
+  @java.beans.ConstructorProperties({"id", "referenceId", "signature", "oscpResponse", "ssn"})
+  public MemberSignedEvent(Long id, String referenceId, String signature, String oscpResponse, String ssn) {
+    this.id = id;
+    this.referenceId = referenceId;
+    this.signature = signature;
+    this.oscpResponse = oscpResponse;
+    this.ssn = ssn;
+  }
+
+  public Long getId() {
```
AlexSJ1990: Why do we get rid of @ Value?

member-service/src/main/java/com/hedvig/memberservice/sagas/MemberSignedSaga.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    @SagaEventHandler(associationProperty = "id")
+    @StartSaga
+    @EndSaga
+    fun onMemberSignedEvent(e: MemberSignedEvent, eventMessage: EventMessage<MemberSignedEvent>) {
+
+        try {
+            underwriterApi.memberSigned(e.getId().toString(), e.getReferenceId(), e.getSignature(), e.getOscpResponse())
+        } catch (ex: RuntimeException) {
+            log.error("Could not notify underwriter about signed member for memberId: {}", e.getId(), ex)
```
AlexSJ1990: shall we do log.error("Could not notify underwriter about signed member for memberId: %, exception is %", e.getId(), ex)

meletios: Correct just fixed it

member-service/src/main/java/com/hedvig/memberservice/events/MemberSignedEvent.java
meletios: Delombok it since we are using this event in Kotlin

PR #357, opened 2020-02-14
Feature/norwegian authentication

FredrikAreschoug:


zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecController.kt
```diff
@@ -0,0 +1,22 @@
...
+@RestController
+@RequestMapping("/hooks/zignsec")
+class ZignSecController(
+    private val zignSecService: ZignSecService
+) {
+
+    @PostMapping(value = ["notifications"])
+    fun webhook(@RequestBody request: ZignSecNotificationRequest): ResponseEntity<String> {
+        zignSecService.handleNotification(request)
+        return ResponseEntity.ok("\uD83D\uDC4D")
```
johantjellden: Are these emojis? 😄

FredrikAreschoug: yes it's a -> :+1:

zign-sec/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+            <artifactId>spring-boot-starter</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-validation</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-data-jpa</artifactId>
+        </dependency>
```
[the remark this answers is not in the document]
FredrikAreschoug: yes maybe we can discuss the pros and cons. If we put it in ZignSec the module could handle everything. If we don't we will have to move in the ZignSec webhook endpoint into member service. An I would like the seperation

PR #359, opened 2020-02-17
Feature/norwegian signing

FredrikAreschoug:


member-service/src/main/java/com/hedvig/integration/underwritter/UnderwriterApi.kt
```diff
@@ -14,4 +16,24 @@ class UnderwriterApi(private val underwriterClient: UnderwriterClient) {
     ) {
         underwriterClient.memberSigned(memberId, SignRequest(referenceToken, signature, oscpResponse))
     }
+
+    fun hasQuoteToSign(memberId: String): QuoteToSignStatusDTO {
+        try {
+            val response = underwriterClient.getQuoteFromMemberId(memberId).body!!
```
meletios: Do we need to check if the member has already signed a quote as well?

member-service/src/main/java/com/hedvig/memberservice/services/NorwegianSigningService.kt
```diff
@@ -0,0 +1,35 @@
...
+            acceptLanguage,
+            request.isMobile
+        )
+        return response.redirectUrl?.let { redirectUrl ->
+            MemberSignResponse(
+                signUUID = response.id,
+                status = SignStatus.IN_PROGRESS,
+                norwegianBankIdResponse = NorwegianBankIdResponse(redirectUrl)
+            )
+        } ?: throw IllegalStateException("Started norwegian sign got no redirect url [Response: $response]")
```
meletios: Add memberId and request for debugging 🙈

member-service/src/main/java/com/hedvig/memberservice/services/NorwegianBankIdService.kt
```diff
@@ -18,14 +18,29 @@ class NorwegianBankIdService(
...
 
         startCollect(response.id)
         return response
     }
 
+    //TODO: maybe we should map more
+    private fun String.toTwoLetterLanguage() = when (this) {
+        "sv-SE" -> "SV"
+        "en-SE" -> "EN"
+        else -> "NO"
```
meletios: What about en-NO?

member-service/src/main/java/com/hedvig/memberservice/services/member/dto/MemberSignResponse.kt
```diff
@@ -0,0 +1,35 @@
...
+
+data class MemberSignResponse(
+    val signId: Long? = null,
+    val signUUID: UUID? = null,
+    val status: SignStatus,
+    val bankIdOrderResponse: OrderResponse? = null,
+    val norwegianBankIdResponse: NorwegianBankIdResponse? = null
+) {
+    init {
+        //assert(!(signId == null && signUUID == null)) { "MemberSignResponse without an id is not valid" }
```
meletios: //TODO: Delete me 😃

member-service/src/main/java/com/hedvig/memberservice/services/member/dto/MemberSignResponse.kt
```diff
@@ -0,0 +1,35 @@
...
+    ): this(
+        signId,
+        null,
+        status,
+        bankIdOrderResponse,
+        null
+    )
+}
+
+data class NorwegianBankIdResponse(
```
meletios: Shall we extract this in to its own class?

member-service/src/main/java/com/hedvig/memberservice/services/SigningService.kt
```diff
@@ -0,0 +1,102 @@
...
+        val existing = signedMemberRepository.findBySsn(request.ssn)
+
+        if (existing.isPresent) {
+            throw MemberHasExistingInsuranceException()
+        }
+
+        val quoteStatus = underwriterApi.hasQuoteToSign(memberId.toString())
+        if (quoteStatus.isEligibleToSign) {
+            val cmd = UpdateWebOnBoardingInfoCommand(memberId, request.ssn, request.email)
+            commandGateway.sendAndWait<Any>(cmd)
```
palmenhq: do we need to pass the generic here? i think its the return value, not the typeof `cmd`

member-service/src/main/java/com/hedvig/memberservice/util/SsnUtilImpl.kt
```diff
@@ -1,6 +1,14 @@
 package com.hedvig.memberservice.util
 
 class SsnUtilImpl : SsnUtil {
+
+    override fun getMarketFromSsn(ssn: String): Market = when (ssn.length) {
+        12 -> Market.SWEDEN
+        11 -> Market.NORWAY
+        else -> throw IllegalArgumentException("Could not get market from ssn [ssn: $ssn]")
+    }
```
palmenhq: this feels shakey. Is there no other way to determine the member's market?

member-service/src/main/java/com/hedvig/memberservice/web/v2/SignController.java
```diff
@@ -16,7 +16,7 @@
 import org.springframework.web.bind.annotation.RestController;
 
 @RestController
-@RequestMapping("/v2/member/sign")
+@RequestMapping("/v2/member/sign/")
```
palmenhq: ?

member-service/src/main/java/com/hedvig/memberservice/services/SigningService.kt
```diff
@@ -0,0 +1,102 @@
...
+    fun IsSsnAlreadySignedMember(ssn: String?): IsSsnAlreadySignedMemberResponse {
+        val existing = signedMemberRepository.findBySsn(ssn)
+        return IsSsnAlreadySignedMemberResponse(existing.isPresent)
+    }
+
+    fun getSignStatus(@NonNull memberId: Long): Optional<SignSession> {
+        val optionalMember = memberRepository.findById(memberId)
+
+        return if (optionalMember.isPresent) {
+            when (SsnUtilImpl.instance.getMarketFromSsn(optionalMember.get().ssn)) {
```
palmenhq: wouldn't it be better to actually check which market the member belongs to with some kind of flag or so?

member-service/src/main/java/com/hedvig/memberservice/services/SigningService.kt
FredrikAreschoug: yes in kotlin forces us to define the return type

member-service/src/main/java/com/hedvig/memberservice/services/SigningService.kt
FredrikAreschoug: I rather that we don't introduce flags if we don't have to. Yes, when checking ssn will not be sufficient for all markets but I rather have us backfill that information than adding flags now. Trying to find out on member data is a more solid solution to me. But yes maybe that's a discussion we should have.

FredrikAreschoug: But we will have a market picker meeting lets see what's the outcome is and let's get back to this

FredrikAreschoug:
@palmenhq There will be a discussion about how to do the market separation beginning of next week, but yes the "shakey" solution that is in this PR will not go out in prod! For the tests I split the tests and will add more for Norway later :) ButIt would be nice to be able to merge this PR into the `feature/norwegian-authentication` soon because I have other PR's relaying on it

FredrikAreschoug:
Nice feedback

PR #363, opened 2020-02-25
removed market and added sign method

FredrikAreschoug:


member-service/src/main/java/com/hedvig/integration/underwriter/dtos/QuoteToSignStatusDto.kt
```diff
@@ -0,0 +1,10 @@
+package com.hedvig.integration.underwriter.dtos
+
+sealed class QuoteToSignStatusDto {
```
meletios: Sweat

PR #364, opened 2020-02-25
Feature/norwegian authentication endpoint

FredrikAreschoug:


member-service/src/main/java/com/hedvig/memberservice/services/NorwegianBankIdService.kt
```diff
@@ -22,6 +33,25 @@ class NorwegianBankIdService(
             )
         )
 
+    fun finishedAuthentication(result: NorwegianAuthenticationResult) {
```
meletios: completeAuthentication maybe?

member-service/src/main/java/com/hedvig/memberservice/web/AuthController.kt
```diff
@@ -40,15 +43,16 @@ import org.springframework.web.bind.annotation.RestController
 import java.io.UnsupportedEncodingException
 
 @RestController
-@RequestMapping("/member/bankid/")
+@RequestMapping("/member/")
```
meletios: Shall we pass the country as a pathvariable and default to sweden if its null?

FredrikAreschoug: sounds crazy! how do you do that? :D

meletios: We can leave the swedish endpoints as you have them actually, but on the Norwegian one, it would some like "/{country}/bankId/authenticate"

meletios: > sounds crazy! how do you do that? :D

Yes it is crazy as i stated it. But check my suggestion up ☝️

PR #368, opened 2020-03-03
Feature/market information on member

akashpatl:
From every bank sign command we also trigger an event for setting the market.

If a member has not yet signed and changed market it will be handled via an endpoint. 

The PR also includes a method for backfilling all existing members with a market that is per default "SE". We'll remove it after use in prod

### DIFF
--- a/member-service/src/main/java/com/hedvig/memberservice/aggregates/Member.java
+++ b/member-service/src/main/java/com/hedvig/memberservice/aggregates/Member.java
@@ -1,6 +1,5 @@
 package com.hedvig.memberservice.aggregates;
 
-import com.hedvig.memberservice.web.dto.Market;
 import lombok.Data;
 
 @Data
--- a/member-service/src/main/java/com/hedvig/memberservice/aggregates/Member.java
+++ b/member-service/src/main/java/com/hedvig/memberservice/aggregates/Member.java
@@ -1,5 +1,6 @@
 package com.hedvig.memberservice.aggregates;
 
+import com.hedvig.memberservice.web.dto.Market;
 import lombok.Data;
 
 @Data
@@ -11,6 +12,8 @@ public class Member {
   private String phoneNumber;
 
   private String acceptLanguage;
+
+  private Market market;
 
   private LivingAddress livingAddress;
 }
--- a/member-service/src/main/java/com/hedvig/memberservice/aggregates/Member.java
+++ b/member-service/src/main/java/com/hedvig/memberservice/aggregates/Member.java
@@ -12,7 +12,7 @@ public class Member {
 
   private String acceptLanguage;
 
-  private Market market;
+  private PickedLocale pickedLocale;
 
   private LivingAddress livingAddress;
 }
--- a/member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
+++ b/member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
@@ -258,8 +258,7 @@ public class MemberAggregate {
     apply(
       new MemberSignedEvent(
         this.id, cmd.getReferenceId(), cmd.getSignature(), cmd.getOscpResponse(),
-        cmd.getPersonalNumber()))
-      .andThenApply(() -> new PickedLocaleUpdatedEvent(this.id, PickedLocale.SE));
+        cmd.getPersonalNumber()));
 
     if (this.trackingId == null) {
       generateTrackingId();
@@ -282,8 +281,7 @@ public class MemberAggregate {
 
     apply(
       new NorwegianMemberSignedEvent(
-        this.id, cmd.getPersonalNumber(), cmd.getProvideJsonResponse()))
-      .andThenApply(() -> new PickedLocaleUpdatedEvent(this.id, PickedLocale.NO));
+        this.id, cmd.getPersonalNumber(), cmd.getProvideJsonResponse()));
 
 
   }
@@ -300,8 +298,7 @@ public class MemberAggregate {
 
   @CommandHandler
   public void on(SignMemberFromUnderwriterCommand signMemberFromUnderwriterCommand) {
-    apply(new MemberSignedWithoutBankId(signMemberFromUnderwriterCommand.getId(), signMemberFromUnderwriterCommand.getSsn()))
-      .andThenApply(() -> new PickedLocaleUpdatedEvent(this.id, PickedLocale.SE));
+    apply(new MemberSignedWithoutBankId(signMemberFromUnderwriterCommand.getId(), signMemberFromUnderwriterCommand.getSsn()));
   }
 
   @CommandHandler
--- a/member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
+++ b/member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
@@ -1,7 +1,5 @@
 package com.hedvig.memberservice.aggregates;
 
-import com.fasterxml.jackson.core.JsonParseException;
-import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.hedvig.common.UUIDGenerator;
 import com.hedvig.external.bisnodeBCI.BisnodeClient;
@@ -80,14 +78,7 @@ public class MemberAggregate {
           return new AcceptLanguageUpdatedEvent(command.getMemberId(), command.getAcceptLanguage());
         }
         return null;
-        })
-      .andThenApply(() ->
-      {
-        if(command.getMarket() != null && !command.getMarket().toString().isEmpty()){
-          return new MarketUpdatedEvent(command.getMemberId(), command.getMarket());
-        }
-        return null;
-      });
+        });
   }
 
   @CommandHandler
@@ -471,7 +462,7 @@ public class MemberAggregate {
 
     if (this.status != MemberStatus.SIGNED &&
       !Objects.equals(member.getMarket(), cmd.getMarket())) {
-      log.info("Updating market for member {}, new locale: {}", cmd.getMemberId(),
+      log.info("Updating market for member {}, new market: {}", cmd.getMemberId(),
         cmd.getMarket());
 
       apply(new MarketUpdatedEvent(cmd.getMemberId(), cmd.getMarket()));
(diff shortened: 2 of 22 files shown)


member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -284,6 +294,9 @@ void norwegianBankIdSignHandler(NorwegianSignCommand cmd) {
     apply(
       new NorwegianMemberSignedEvent(
         this.id, cmd.getPersonalNumber(), cmd.getProvideJsonResponse()));
+
+    apply(new MarketUpdatedEvent(this.id, Market.NO));
+
```
palmenhq: should this maybe be `andThenApply`? or does the order of these commands make a difference?

member-service/src/main/java/com/hedvig/memberservice/query/MemberEntity.java
```diff
@@ -50,6 +51,9 @@
 
   public String acceptLanguage;
 
+  @Enumerated(EnumType.STRING)
+  public Market market;
```
palmenhq: Maybe this could be marked with `@Nullable`?

member-service/src/main/java/com/hedvig/memberservice/web/InternalMembersController.java
```diff
@@ -208,5 +190,16 @@ public void editMember(
     commandBus.sendAndWait(new UpdateSSNCommand(memberId, request.getSsn()));
     return ResponseEntity.noContent().build();
   }
+
+  //TODO: THIS IS A ONE OFF TO BACKFILL MARKET. REMOVE AFTER USE!
+  @PostMapping(value = "/backfill/market")
+  public ResponseEntity<Void> backfillMarket(
```
palmenhq: This feels dangerous, should we add some kind of mechanism for making it run once only? or is that unnecessary?

member-service/src/main/java/com/hedvig/memberservice/web/InternalMembersController.java
```diff
@@ -208,5 +190,16 @@ public void editMember(
...
+  ) {
+    memberRepository
+      .findAll()
+      .forEach(m ->
+        commandBus.sendAndWait(new BackfillMarketCommand(m.getId())));
```
palmenhq: Could it be an idea to do this in bulks? I'm thinking since we have a lot of members? My gut feeling is that loading hundreds of thousands of members could be memory hungry.

member-service/src/main/java/com/hedvig/memberservice/web/InternalMembersController.java
johantjellden: We could have a custom command for this and make sure that the aggregate only emit the event one time.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
akashpatl: Good point! yes they do, we need a sign before applying the market.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -79,7 +80,14 @@ public MemberAggregate(CreateMemberCommand command) {
           return new AcceptLanguageUpdatedEvent(command.getMemberId(), command.getAcceptLanguage());
         }
         return null;
-        });
+        })
+      .andThenApply(() ->
+      {
+        if(command.getMarket() != null && !command.getMarket().toString().isEmpty()){
```
vonElfvin: Why would the string be empty in an enum?
```suggestion
        if(command.getMarket() != null){
```

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -299,6 +312,8 @@ public boolean isValidJSON(final String json) {
   @CommandHandler
   public void on(SignMemberFromUnderwriterCommand signMemberFromUnderwriterCommand) {
     apply(new MemberSignedWithoutBankId(signMemberFromUnderwriterCommand.getId(), signMemberFromUnderwriterCommand.getSsn()));
+
+    apply(new MarketUpdatedEvent(this.id, Market.SE));
```
vonElfvin: How come Market.SE is hardcoded here?

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -453,6 +468,26 @@ public void on(UpdateAcceptLanguageCommand cmd) {
     }
   }
 
+  @CommandHandler
+  public void on(UpdateMarketCommand cmd) {
+
+    if (this.status != MemberStatus.SIGNED &&
+      !cmd.getMarket().toString().isEmpty() &&
```
vonElfvin: Seems unlikely once agian that a enum would be empty as a string

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -453,6 +468,26 @@ public void on(UpdateAcceptLanguageCommand cmd) {
...
+      !Objects.equals(member.getMarket(), cmd.getMarket())) {
+      log.info("Updating accept locale for member {}, new locale: {}", cmd.getMemberId(),
```
vonElfvin: Updating accept locale?

member-service/src/main/java/com/hedvig/memberservice/commands/CreateMemberCommand.kt
```diff
@@ -1,9 +1,11 @@
...
 
+import com.hedvig.memberservice.web.dto.Market
 import org.axonframework.commandhandling.TargetAggregateIdentifier
 
 data class CreateMemberCommand(
     @TargetAggregateIdentifier
     val memberId: Long,
-    val acceptLanguage: String? = null
+    val acceptLanguage: String? = null,
+    val market: Market? = null
```
vonElfvin: I would prefer is this was not nullable tbh

member-service/src/main/java/com/hedvig/memberservice/web/MembersController.java
```diff
@@ -160,4 +159,12 @@ public MembersController(MemberRepository repo,
 
         return ResponseEntity.accepted().build();
     }
+
+  @PostMapping("/locale/update")
+  public ResponseEntity<?> postLanguage(@RequestHeader(value = "hedvig.token") Long hid, @RequestBody @Valid PostLocaleRequestDTO body) {
+
+    commandGateway.sendAndWait(new UpdateMarketCommand(hid, body.getMarket()));
```
vonElfvin: locale != market?

member-service/src/main/java/com/hedvig/memberservice/web/v2/MembersControllerV2.kt
```diff
@@ -40,7 +40,7 @@ class MembersControllerV2 @Autowired constructor(
                 member = memberRepository.findById(memberId)
             } while (member.isPresent)
 
-            commandGateway.send<Void>(CreateMemberCommand(memberId!!, acceptLanguage))
+            commandGateway.send<Void>(CreateMemberCommand(memberId!!, acceptLanguage, null))
```
vonElfvin: If we want to close this "data" hole going forward we should not send null, rather send the market as a nullbable in the DTO and default to SE if it is null.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
akashpatl: @FredrikAreschoug has assured me that "signs without BankID" will always be for sweden.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
akashpatl: 👍  Legacy from when it was a string and not enum.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
akashpatl: 👍

member-service/src/main/java/com/hedvig/memberservice/commands/CreateMemberCommand.kt
akashpatl: Might actually not even need to have market on the CreateMemberCommand. It's only used for Apple

member-service/src/main/java/com/hedvig/memberservice/web/MembersController.java
akashpatl: 👍

member-service/src/main/java/com/hedvig/memberservice/web/InternalMembersController.java
akashpatl: I have a custom command called BackfillMarketCommand that we can remove after use.

member-service/src/main/java/com/hedvig/memberservice/web/v2/MembersControllerV2.kt
akashpatl: Removed market from the DTO. We don't really need it there and it's also nullable.

akashpatl:
I introduced a batch process for the backfill that reads 1000 ids in batches and triggers the update market events. The batch reads from the databse is quite fast but the event application is a bit slow (ca 1 hour for 17K ids on my local machine). We'll need to test it in staging.

member-service/src/main/java/com/hedvig/memberservice/aggregates/Member.java
```diff
@@ -1,5 +1,6 @@
 package com.hedvig.memberservice.aggregates;
 
+import com.hedvig.memberservice.web.dto.Market;
```
johantjellden: Move this enum into the aggregates package. I think it it cleaner if the web package references the aggregates then the other way around.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -453,6 +457,26 @@ public void on(UpdateAcceptLanguageCommand cmd) {
     }
   }
 
+  @CommandHandler
+  public void on(UpdateMarketCommand cmd) {
+
+    if (this.status != MemberStatus.SIGNED &&
```
johantjellden: Maybe we should do this for all aggreages not only the ones that are not singed?

member-service/src/main/java/com/hedvig/memberservice/aggregates/Member.java
akashpatl: 👍

akashpatl:
We've decided to call this picked_locale and it will only be updated via a graphql mutation and not based on signing.

PR #372, opened 2020-03-11
replaced RuntimeException with new MemberNotFoundException

ipinak:


member-service/src/main/java/com/hedvig/memberservice/aggregates/exceptions/MemberNotFoundException.java
```diff
@@ -0,0 +1,7 @@
+package com.hedvig.memberservice.aggregates.exceptions;
+
+public class MemberNotFoundException extends Exception {
```
FredrikAreschoug:
```suggestion
public class MemberNotFoundException extends RuntimeException {
```

FredrikAreschoug: will probably also fix the tests :)

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -188,7 +189,7 @@ void inactivateMember(InactivateMemberCommand command) {
       apply(new MemberInactivatedEvent(this.id));
     } else {
       String str =
-        String.format("Cannot INACTIAVTE member %s in status: %s", this.id, this.status.name());
+        String.format("Cannot DEACTIVATE member %s in status: %s", this.id, this.status.name());
```
FredrikAreschoug: Just following the event naming, but yes it should be `MemberDeactivatedEvent`

ipinak:
As I thought this is not the problem :(

PR #376, opened 2020-03-26
update pickedLocale enum type to match Locale graphqltype from giraffe

akashpatl:


member-service/src/main/java/com/hedvig/memberservice/web/MembersController.java
```diff
@@ -161,10 +161,17 @@ public MembersController(MemberRepository repo,
...
+  public ResponseEntity<Member> postPickedLocale(@RequestHeader(value = "hedvig.token") Long hid, @RequestBody @Valid PostPickedLocaleRequestDTO body) {
 
     commandGateway.sendAndWait(new UpdatePickedLocaleCommand(hid, body.getPickedLocale()));
 
-    return ResponseEntity.accepted().build();
+    Optional<MemberEntity> member = repo.findById(hid);
+    if (member.isPresent()) {
+
+      return ResponseEntity.ok(new Member(member.get()));
+    }
```
meletios: do we need to return the Member here? or just accepted?

member-service/src/main/java/com/hedvig/memberservice/web/dto/PickedLocaleDTO.java
```diff
@@ -0,0 +1,18 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import lombok.Value;
+
+import java.util.List;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
```
meletios: You dont need these

member-service/src/main/java/com/hedvig/memberservice/web/InternalMembersController.java
```diff
@@ -210,5 +211,21 @@ public void editMember(
...
   }
+
+  @GetMapping("/{memberId}/pickedLocale")
+  public ResponseEntity<PickedLocaleDTO> getPickedLocale(@PathVariable Long memberId) {
+
+    Optional<MemberEntity> member = memberRepository.findById(memberId);
+    if (member.isPresent() && member.get().pickedLocale != null) {
+
+      PickedLocaleDTO res = new PickedLocaleDTO();
+      res.setPickedLocale(member.get().pickedLocale);
```
meletios: You can pass it as a parameter to the constructor

member-service/src/main/java/com/hedvig/memberservice/web/dto/PickedLocaleDTO.java
meletios: if you apply my suggestion below eh

member-service/src/main/java/com/hedvig/memberservice/web/MembersController.java
akashpatl: I've had a discussion with @hedvigoscar and he insists we should return it. Since he is the consumer of data I think he should decide.

member-service/src/main/java/com/hedvig/memberservice/web/dto/PickedLocaleDTO.java
akashpatl: Habit. Will remove

member-service/src/main/java/com/hedvig/memberservice/web/InternalMembersController.java
akashpatl: 👍

akashpatl: Need to keep the Lombok annotations though ;)

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -470,7 +470,7 @@ public void on(UpdatePickedLocaleCommand cmd) {
   @CommandHandler
   public void on(BackfillPickedLocaleCommand cmd) {
 
-      apply(new PickedLocaleUpdatedEvent(cmd.getMemberId(), PickedLocale.SE));
+      apply(new PickedLocaleUpdatedEvent(cmd.getMemberId(), PickedLocale.sv_SE));
```
johantjellden: Maybe we should check here if locale is already picked so that we could run this many times.

akashpatl: we already do this in the controller when we trigger the BackfillPickedLocaleCommand but I can add an extra check. I agree with that the logic for the check should be here.

PR #381, opened 2020-03-30
fix: new session on new personal number and feature: success and fail url from uw

FredrikAreschoug:


member-service/src/main/java/com/hedvig/memberservice/services/NorwegianBankIdService.kt
```diff
@@ -26,22 +26,26 @@ class NorwegianBankIdService(
...
-                resolveTwoLetterLanguageFromMember(memberId)
+                resolveTwoLetterLanguageFromMember(memberId),
+                request.successUrl,
+                request.failUrl
             )
         )
     }
 
-    fun sign(memberId: String, ssn: String) =
+    fun sign(memberId: String, ssn: String, targetUrl: String, failedTargetUrl: String) =
```
meletios: shall we rename them to successUrl and failUrl ? just to follow the same naming

PR #382, opened 2020-03-31
Feature/notification service customerio

johantjellden:


member-service/src/main/java/com/hedvig/memberservice/services/segmentpublisher/EventListener.kt
```diff
@@ -29,18 +28,17 @@ class EventListener @Autowired constructor(
     @EventHandler
     fun on(evt: EmailUpdatedEvent) {
         val traits = ImmutableMap
-            .of<String, Any>("email", evt.email,
-                "timezone", "Europe/Stockholm ")
+            .of<String, Any>(
+                "email", evt.email,
+                "timezone", "Europe/Stockholm "
```
meletios, line 33: Is this true for everyone though?

johantjellden: Good catch

meletios: 🦅 👁

PR #383, opened 2020-04-01
Feature/set default picked locale

akashpatl:


member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -458,13 +458,8 @@ public void on(UpdateAcceptLanguageCommand cmd) {
...
   public void on(UpdatePickedLocaleCommand cmd) {
-
-    if (!Objects.equals(member.getPickedLocale(), cmd.getPickedLocale())) {
-      log.info("Updating picked locale for member {}, new locale: {}", cmd.getMemberId(),
-        cmd.getPickedLocale());
-
-      apply(new PickedLocaleUpdatedEvent(cmd.getMemberId(), cmd.getPickedLocale()));
-    }
+    log.info("Updating picked locale for member {}, new locale: {}", cmd.getMemberId(), cmd.getPickedLocale());
+    apply(new PickedLocaleUpdatedEvent(cmd.getMemberId(), cmd.getPickedLocale()));
```
vonElfvin, line 462: Why remove the "sanity check", no point in sending the "PickedLocaleUpdatedEvent" when in reality it wasn't I'd say

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -458,13 +458,8 @@ public void on(UpdateAcceptLanguageCommand cmd) {
 
   @CommandHandler
   public void on(UpdatePickedLocaleCommand cmd) {
-
-    if (!Objects.equals(member.getPickedLocale(), cmd.getPickedLocale())) {
```
meletios, line 462: we need this eller? We dont need to change the locale if its the same

meletios:
Do we need to default it in the db also?

johantjellden:
@FredrikAreschoug We can close this right?

FredrikAreschoug:
yes lets just do it!

PR #386, opened 2020-04-02
fail if member changes ssn during sign

FredrikAreschoug:


zign-sec/src/test/java/com/hedvig/external/zignSec/ZignSecSessionServiceImplTest.kt
```diff
@@ -317,7 +317,7 @@ class ZignSecSessionServiceImplTest {
             notification = null,
             createdAt = timestamp,
             updatedAt = timestamp,
-            requestPersonalNumber = null
+            requestPersonalNumber = ""
```
vonElfvin, line 320: Hahah the good ol'

PR #387, opened 2020-04-03
Feature/add cash back option norway

FredrikAreschoug:
:norway: :moneybag:


member-service/src/main/java/com/hedvig/memberservice/web/CashbackController.java
```diff
@@ -61,7 +62,7 @@ public CashbackController(
       return ResponseEntity.ok(cashbackOptions);
     }
 
-    return ResponseEntity.ok(cashbackService.getOptions());
+    return ResponseEntity.ok(cashbackService.getOptions(PickedLocale.sv_SE));
```
johantjellden, line 65: Should we default to a dummy option that only have empty data?

PR #396, opened 2020-04-14
Feature/zign sec use id provider personal

FredrikAreschoug:


zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
```diff
@@ -206,18 +202,62 @@ class ZignSecSessionServiceImpl(
...
+                        if (signEntity.isPresent) {
+                            norwegianAuthenticationEventPublisher.publishAuthenticationEvent(
+                                NorwegianAuthenticationResult.Completed(
+                                    session.referenceId,
+                                    session.memberId,
+                                    signEntity.get().personalNumber
+                                )
+                            )
+                        } else {
+                            logger.error("Member tried to login whit no ZignSecSignEntity [MemberId:${session.memberId}] [SessionId:${session.sessionId}] [session:$session]")
```
meletios, line 219: Do we need `idProviderPersonId` and `requestPersonalNumber` on this? It will be helpful i think

FredrikAreschoug: requestPersonalNumber will always be null and idProviderPersonId should be included in session

PR #399, opened 2020-04-30
endpoint to see if member is signed

FredrikAreschoug:


member-service/src/main/java/com/hedvig/memberservice/services/SigningService.kt
```diff
@@ -85,11 +86,16 @@ class SigningService(
         }
     }
 
-    fun IsSsnAlreadySignedMember(ssn: String?): IsSsnAlreadySignedMemberResponse {
+    fun isSsnAlreadySignedMember(ssn: String?): IsSsnAlreadySignedMemberResponse {
         val existing = signedMemberRepository.findBySsn(ssn)
         return IsSsnAlreadySignedMemberResponse(existing.isPresent)
     }
 
+    fun isMemberAlreadySigned(memberId: Long): IsMemberAlreadySignedResponse {
```
meletios, line 94: Can we have both ssn or memberId? Just to not call two times member-service?

FredrikAreschoug: I was thinking that first but we look if a member has signed after they went thought the bankid. But for this, I think we should check in start sign. That's why I split them up. You could look at this one to see my plan -> https://github.com/HedvigInsurance/underwriter/pull/107

PR #403, opened 2020-07-02
Add endpoint to check if a person has signed with same ssn/email

vonElfvin:


member-service/src/main/java/com/hedvig/personservice/persons/web/PersonController.kt
```diff
@@ -79,4 +80,10 @@ class PersonController @Autowired constructor(
         val personFlags = personService.getAllPersonFlags(person)
         return ResponseEntity.ok(personFlags)
     }
+
+    @PostMapping("/has/signed")
```
FredrikAreschoug, line 84: I think this endpoint is more descriptive
```suggestion
    @PostMapping("/hasPersonSigned")
```

FredrikAreschoug: or 
```suggestion
    @PostMapping("/hasPersonBeenSigned")
```

vonElfvin: Since it is "/person" on the top I felt like this was pretty neat,  maybe "hasSigned"?

FredrikAreschoug: Totaly missed that sorry about that

PR #405, opened 2020-07-24
Add QualityAssuranceService, available only in staging and development

rasmusguterstam:
### What's up

Adds a QualityAssuranceService to be used for QA-related-endpoints. 

Currently only implements `unsignMember`.

Btw, would love to get some feedback on appropriate naming of the route `/_/staging/{country}/member/unsign`, I just went with something 🤓


member-service/src/main/java/com/hedvig/memberservice/web/QualityAssuranceController.kt
```diff
@@ -0,0 +1,25 @@
...
+
+@RestController
+@Profile("staging", "development")
+@RequestMapping("/_/staging/")
+class QualityAssuranceController(
+    private val qualityAssuranceService: QualityAssuranceServiceImpl
+) {
+    @PostMapping("{country}/member/unsign")
+    fun unsignMember(
+        @PathVariable("country") country: String,
```
FredrikAreschoug, line 17: We usually call it market, so I would go with that instead of country. Another neet thing is that you can have an enum as the path variable which is quite neat because then you don't need to check and throw `CountryNotFoundException` hence you will get a 404 since the path doesn't exist.

member-service/src/main/java/com/hedvig/memberservice/services/QualityAssuranceServiceImpl.kt
```diff
@@ -0,0 +1,52 @@
...
+            "NORWAY" -> unsignNorwegianMember(ssn)
+            else -> throw CountryNotFoundException(country)
+        }
+
+    private fun unsignSwedishMember(ssn: String): Boolean {
+        val signedMemberEntity = signedMemberRepository.findBySsn(ssn).orNull()
+            ?: throw SignedMemberNotFoundException(
+                ssn = ssn,
+                repositoryName = "SignedMemberRepository"
+            )
```
FredrikAreschoug, line 29: I like your plan here but optional actually have a `orElseThrow` which you could use like this: 
```suggestion
        val signedMemberEntity = signedMemberRepository.findBySsn(ssn).orElseThrow {
            SignedMemberNotFoundException(
                ssn = ssn,
                repositoryName = "SignedMemberRepository"
            )
            }
```

member-service/src/main/java/com/hedvig/memberservice/web/QualityAssuranceController.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.*
+
+@RestController
+@Profile("staging", "development")
+@RequestMapping("/_/staging/")
+class QualityAssuranceController(
+    private val qualityAssuranceService: QualityAssuranceServiceImpl
+) {
+    @PostMapping("{country}/member/unsign")
```
FredrikAreschoug, line 15: Could be since it's one action
```suggestion
    @PostMapping("{country}/unsignMember")
```

member-service/src/main/java/com/hedvig/memberservice/web/QualityAssuranceController.kt
rasmusguterstam: Ooh that's really neat actually!

member-service/src/main/java/com/hedvig/memberservice/util/OptionalExtension.kt
```diff
@@ -0,0 +1,5 @@
+package com.hedvig.memberservice.util
+
+import java.util.Optional
+
+fun <T> Optional<T>.orNull(): T? = orElse(null)
```
FredrikAreschoug, line 5: remove it since it is not used any more :dagger:

PR #408, opened 2020-08-20
Datadog monitoring

palmenhq:
An attempt to gather application-specific metrics in datadog, so we can actually monitor and set up alerts if we find anomalities. This PR covers bank id signs and auth attempts and results, but we could extend it to for example connected payments and # of messages flowing through the system. See [DogStatsD docs](https://docs.datadoghq.com/developers/dogstatsd/?tab=hostagent#how-it-works) for implementation details, if you're interested :) I'm not sure this is production ready, i'd like to go through it some more and maybe write some tests. BUT I'd like your input before i spend more time on it ⌛ 

Bonus: bankid sign/auth refactoring. Needs extremely thorough testing obviously, as i've been unable to test the full flow manually. Actually, i'm thinking maybe some of this is even deprecated and could be removed? I've found like four places or something where we trigger bankid requests so i guess some of them are outdated? plx help


member-service/src/main/java/com/hedvig/memberservice/services/bankid/BankIdAuthCollectionResult.kt
```diff
@@ -0,0 +1,19 @@
+package com.hedvig.memberservice.services.bankid
+
+import com.hedvig.memberservice.web.dto.BankIdProgressStatus
+
+class BankIdAuthCollectionResult(
```
hedvigoscar, line 5: Could this be expressed as a sealed class? Currently, it's possible to have a `BankIdAuthCollectionResult` that is `.success==null`,`.error==null` and `.status==null`.

member-service/src/main/java/com/hedvig/memberservice/services/bankid/BankIdService.kt
```diff
@@ -0,0 +1,156 @@
...
+@Service
+class BankIdService(
+    private val bankIdApi: BankIdApi,
+    private val collectRepository: CollectRepository,
+    private val commandGateway: CommandGateway,
+    private val memberRepository: MemberRepository,
+    private val signedMemberRepository: SignedMemberRepository,
+    private val metricsCollector: MetricsCollector
+) {
+    private val log = LoggerFactory.getLogger(BankIdService::class.java)
```
hedvigoscar, line 35: I wrote a really nifty Kotlin construct that allows you to in a neat way add loggers to anything, it looks like this:

```inline fun <reified T> T.logger() = lazy { LoggerFactory.getLogger(T::class.java)```

which allows you to do this:

```private val log by logger()```

could be interesting to use 🙂

File: member-service/src/main/java/com/hedvig/memberservice/services/metrics/MetricsCollector.kt [hunk 3]
```diff
@@ -0,0 +1,12 @@
+package com.hedvig.memberservice.services.metrics
+
+interface MetricsCollector {
```
hedvigoscar, line 3: Could it be interesting to have a `SwedishBankIdMetricsCollector`? That way, Norwegian BankID does not have to bother with the methods from the Swedish BankID and vice versa 🙂 

Also, is this interface used for anything but having one implementation? If not, it could probably be removed 😄

member-service/src/main/java/com/hedvig/memberservice/services/metrics/MetricsCollectorImpl.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory.getLogger
+import org.springframework.stereotype.Service
+
+@Service
+class MetricsCollectorImpl(
+    private val dogStatsDClient: StatsDClient
+) : MetricsCollector {
+    companion object {
+        val logger: Logger = getLogger(MetricsCollectorImpl::class.java)
```
hedvigoscar, line 14: My nifty inline function could be used here as well 😬

member-service/src/main/java/com/hedvig/memberservice/services/metrics/MetricsCollectorImpl.kt
```diff
@@ -0,0 +1,76 @@
...
+            event,
+            when (result) {
+                ExecutionResult.Success -> "status:success"
+                ExecutionResult.Error -> "status:error"
+            }
+        )
+    }
+
+    private fun registerEvent(title: String) {
+        logger.info("Sending event \"$title\" to DogStatsD")
```
hedvigoscar, line 70: If we worry even a little bit about exceptions in the DataDog-internals, we could wrap all this in `runCatching` as we're kind of fine with discarding the errors 🙂

member-service/src/main/java/com/hedvig/memberservice/services/metrics/MetricsCollector.kt
hedvigoscar: I actually noticed that we're using it in a test.

I think `MockK` can mock arbitrary classes though, without requiring splitting it out to an interface with one implementation 🙂

hedvigoscar (commented):
It could be useful to split out the refactoring parts to their own PR, that usually makes for more easily consumed PRs 🙂

member-service/src/main/java/com/hedvig/memberservice/services/bankid/BankIdService.kt
hedvigoscar: Come to think of it, it does not even have to be `lazy`, in fact it's probably smarter just to have it be

```kotlin
inline fun <reified T> T.logger() = LoggerFactory.getLogger(T::class.java)
```

hedvigoscar: Then again you couldn't do `by logger()` in that case since it would not be a delegate any more 🤔

hedvigoscar: Who cares, really. It's in Backend, it's trivial bikeshedding 😬

member-service/src/main/java/com/hedvig/memberservice/services/metrics/MetricsCollector.kt
palmenhq: Generally we do this for all services in the backend, i agree it's a bit chatty but i think we should keep following that convention.

hedvigoscar: I think this might actually be a good example of Cargo Cult-programming 😉

palmenhq:
Actually, reading up on this i don't think collecting events is the right case, after reading the [Monitoring 101](https://www.datadoghq.com/blog/monitoring-101-collecting-data/) blog post from Datadog. 

> some monitoring systems can also capture events: discrete, infrequent occurrences that can provide crucial context for understanding what changed in your system’s behavior.

Back to the drawing board...

PR #409, opened 2020-08-31
Trim firstName and lastName when searching for members

vonElfvin:
Aparently some members have firstNames like "Elvin " instead of "Elvin"


AlexSJ1990 (approved):
✅

meletios (approved):
This will only solve the searching right?

vonElfvin:
> This will only solve the searching right?

That is correct. We should solve the trimming of firstNames as well :)

PR #418, opened 2020-10-29
gender info on market

FredrikAreschoug:
This was exploding in [sentry](https://sentry.io/organizations/hedvig/issues/1958964110/?environment=staging&project=296775&query=is%3Aunresolved) so I did a 'proper' fix. This will not scale but it will work for now :slightly_smiling_face:

### DIFF
--- a/member-service/src/main/java/com/hedvig/memberservice/util/SsnUtilImpl.kt
+++ b/member-service/src/main/java/com/hedvig/memberservice/util/SsnUtilImpl.kt
@@ -7,10 +7,28 @@ class SsnUtilImpl : SsnUtil {
             return null
         }
 
-        return if ((ssn.replace("-", "").toCharArray()[10].toInt() % 2) == 0) {
-            Gender.FEMALE
+        val trimmedSSN = ssn.replace("-", "")
+        return if (trimmedSSN.length == 10) {
+            //DANISH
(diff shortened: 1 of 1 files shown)


member-service/src/main/java/com/hedvig/memberservice/util/SsnUtilImpl.kt
```diff
@@ -7,10 +7,34 @@ class SsnUtilImpl : SsnUtil {
             return null
         }
 
-        return if ((ssn.replace("-", "").toCharArray()[10].toInt() % 2) == 0) {
-            Gender.FEMALE
-        } else {
-            Gender.MALE
+        val trimmedSSN = ssn.replace("-", "")
```
vonElfvin, line 10:
```suggestion
        val trimmedSSN = ssn.replace("-", "")
        val market = when (trimmedSSN.length) {
          10 -> Market.DENMARK
          ...
        }
        return when (market) ...
```

vonElfvin (approved):
Left a comment without a review before 😄 This works but a small suggestion to make the code a bit prettier 😸

FredrikAreschoug:
Yeah good suggestion but I don't want to create yet another `Market` enum for now, so I will skip it :wink:

PR #421, opened 2020-11-04
HVG-317: Localize Cashbacks

hedvigoscar:


member-service/src/main/java/com/hedvig/memberservice/web/CashbackController.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.*
+import java.util.*
+
+@RestController
+@RequestMapping("/cashback")
+class CashbackController @Autowired constructor(
+    var commandGateway: CommandGateway,
+    val memberRepository: MemberRepository,
```
vonElfvin, line 18:
```suggestion
    private val commandGateway: CommandGateway,
    private val memberRepository: MemberRepository,
```

member-service/src/main/java/com/hedvig/memberservice/web/CashbackController.kt
```diff
@@ -0,0 +1,58 @@
...
+    private val cashbackService: CashbackService
+) {
+    @GetMapping("options")
+    fun options(
+        @RequestHeader(value = "hedvig.token") hid: Long
```
vonElfvin, line 23:
```suggestion
        @RequestHeader(value = "hedvig.token") memberId: Long
```

member-service/src/main/java/com/hedvig/memberservice/web/dto/CashbackOption.kt
```diff
@@ -0,0 +1,53 @@
...
+"id": "someotherid",
+"title": "Mitt konto",
+"description": "Lorem ipsum dolor sit amet...",
+"selected": true,
+"charity": false,
+"imageUrl": "https://unsplash.it/400/200"
+}
+]*/
+data class CashbackOption(
+    @JvmField
```
vonElfvin, line 25: Do you need @JvmField

member-service/src/main/java/com/hedvig/memberservice/services/cashback/LocalizedCashbackService.kt
```diff
@@ -0,0 +1,111 @@
...
+
+@Component
+@ConditionalOnProperty(name = ["localizedCashback"], havingValue = "true")
+class LocalizedCashbackService(
+    private val localizationService: LocalizationService
+) : CashbackService {
+
+    init {
+        println()
+    }
```
vonElfvin, line 26: What

member-service/src/main/java/com/hedvig/memberservice/services/cashback/LocalizedCashbackService.kt
```diff
@@ -0,0 +1,111 @@
...
+            if (cashbackOption != null) return Optional.ofNullable(localize(cashbackOption, pickedLocale.locale))
+        } catch (ex: NullPointerException) {
+        }
+        return Optional.ofNullable(swedishOptions[cashbackId]?.let { localize(it, pickedLocale.locale) })
+    }
+
+    override fun getOptions(pickedLocale: PickedLocale) =
+        if (isNorwegian(pickedLocale)) {
+            norwegianOptions.values.toList().map { localize(it, pickedLocale.locale) }
+        } else swedishOptions.values.toList().map { localize(it, pickedLocale.locale) }
```
vonElfvin, line 40: Knowing we're heading to Denmark maybe this can be more like 
```
when {
  isNorwegian ->
  isSwedish
  isDanish  ->
  else -> throw
```

member-service/src/main/java/com/hedvig/memberservice/services/cashback/LocalizedCashbackService.kt
```diff
@@ -0,0 +1,111 @@
...
+        localizationService.getTranslation(nonLocalized.nameKey, locale),
+        localizationService.getTranslation(nonLocalized.titleKey, locale),
+        localizationService.getTranslation(nonLocalized.descriptionKey,locale),
+        nonLocalized.selected,
+        nonLocalized.charity,
+        nonLocalized.imageUrl,
+        nonLocalized.selectedUrl,
+        localizationService.getTranslation(nonLocalized.signatureKey, locale),
+        localizationService.getTranslation(nonLocalized.paragraphKey, locale)
+    )
```
vonElfvin, line 62: Is it possible to add the names here?
```suggestion
    private fun localize(nonLocalized: NonLocalizedCashbackOption, locale: Locale) = CashbackOption(
        id = nonLocalized.id,
        ...
    )
```

vonElfvin: If it's Java it won't work of course

vonElfvin (commented):
Good job! Great ownership! Added some thoughts :)

member-service/src/main/java/com/hedvig/memberservice/services/cashback/LocalizedCashbackService.kt
```diff
@@ -0,0 +1,111 @@
...
+        } catch (ex: NullPointerException) {
+        }
+        return Optional.ofNullable(swedishOptions[cashbackId]?.let { localize(it, pickedLocale.locale) })
+    }
+
+    override fun getOptions(pickedLocale: PickedLocale) =
+        if (isNorwegian(pickedLocale)) {
+            norwegianOptions.values.toList().map { localize(it, pickedLocale.locale) }
+        } else swedishOptions.values.toList().map { localize(it, pickedLocale.locale) }
+
```
FredrikAreschoug, line 41: Look over the semantics here `getDefaultId` looks better to me

FredrikAreschoug (commented):
Feels like this became a bit complex, do we need to keep `OldCashbackService`. Could we look at the option vals they are hardcoded but we don't always use them for getting the UUID's

member-service/src/main/java/com/hedvig/memberservice/web/dto/CashbackOption.kt
hedvigoscar: Yes, due to this file being used from some old Java-classes

member-service/src/main/resources/application-staging.yml
```diff
@@ -0,0 +1 @@
+localizedCashback: true
```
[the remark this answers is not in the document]
hedvigoscar: We should have both, but I suppose config maps might be the best strategy? 🤔 @johantjellden thoughts?

member-service/src/main/java/com/hedvig/memberservice/services/cashback/LocalizedCashbackService.kt
```diff
@@ -0,0 +1,111 @@
...
+    init {
+        println()
+    }
+
+    override fun getCashbackOption(cashbackId: UUID?, pickedLocale: PickedLocale): Optional<CashbackOption> {
+        try {
+            val cashbackOption = norwegianOptions[cashbackId]
+            if (cashbackOption != null) return Optional.ofNullable(localize(cashbackOption, pickedLocale.locale))
+        } catch (ex: NullPointerException) {
+        }
```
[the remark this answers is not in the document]
hedvigoscar: I think it's a quirk of auto-converting the file to Kotlin. I'll remove it!

member-service/src/main/java/com/hedvig/memberservice/services/cashback/LocalizedCashbackService.kt
```diff
@@ -0,0 +1,111 @@
...
+        nonLocalized.selected,
+        nonLocalized.charity,
+        nonLocalized.imageUrl,
+        nonLocalized.selectedUrl,
+        localizationService.getTranslation(nonLocalized.signatureKey, locale),
+        localizationService.getTranslation(nonLocalized.paragraphKey, locale)
+    )
+
+    private fun isNorwegian(pickedLocale: PickedLocale): Boolean {
+        return pickedLocale === PickedLocale.nb_NO || pickedLocale === PickedLocale.en_NO
```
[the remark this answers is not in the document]
hedvigoscar: Kind of - I don't have the content that is appropriate for Denmark, though.

hedvigoscar:
We keep it only for deployment purposes, it is to be deleted after deployment.

Re: bringing this up to shape for Denmark, I'm not actually sure we should invest too much into this code - this code looks to me a lot like it should actually live in `content-service`.

hedvigoscar:
As such, I did not spend too much time bringing the code generally up to spec, preferring instead to convert it to a working localized state - most of the comments are really on the code as it existed before 🤔

member-service/src/main/java/com/hedvig/common/localization/LocalizationServiceImpl.kt
```diff
@@ -0,0 +1,20 @@
+package com.hedvig.common.localization
+
+import com.hedvig.lokalise.client.LokaliseClient
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Component
+import java.util.*
+
+@Component
+class LocalizationServiceImplForReal(
```
FredrikAreschoug, line 10: Drop the ForReal

FredrikAreschoug:
I defiantly agree that most of this should be in content service and yes we will have to re-do it for Denmark. But I don't see why we should have duplicated code in the mean time.

member-service/src/main/java/com/hedvig/common/localization/LocalizationServiceImpl.kt
hedvigoscar: If I do then it clashes with a registered bean from the old localization package, which we for other reasons apparently cannot remove

member-service/src/main/resources/application-staging.yml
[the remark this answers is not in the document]
johantjellden: I think that it is ok to have this in the source tree. Since this file will only be loaded when the stagning profile is active. And you can override it by setting it in the configmap.

member-service/src/main/java/com/hedvig/common/localization/LocalizationServiceImpl.kt
FredrikAreschoug: Why do we then need both?

FredrikAreschoug: Aha now I get it do we use the old package? cant we remove that package? the new lokalisation package should have everything you need to remove the old one

hedvigoscar: I'll check it out! It's used to resolve some locale-related think iirc

hedvigoscar: You were right - this is now fixed

johantjellden (approved):

PR #435, opened 2020-12-07
Check if the member has signed when checking if a person has signed

vonElfvin:
Why?
-  We have a bug where we update the customer.io attributes (first name, last name, email + send event) for members who onboard again on the same device but with new info

What?
- Check if the member has signed as well as the ssn and email

Risks?
- We have to deploy the consumers of this endpoint at the same time since the dto was extended with new non-nullable properties. I could do it by creating another endpoint etc. but I figured the risk is low enough to not bother with that.

### DIFF
--- a/member-service/src/main/java/com/hedvig/personservice/persons/web/PersonController.kt
+++ b/member-service/src/main/java/com/hedvig/personservice/persons/web/PersonController.kt
@@ -1,13 +1,19 @@
 package com.hedvig.personservice.persons.web
 
-import com.hedvig.personservice.persons.web.dtos.HasPersonSignedBeforeRequest
 import com.hedvig.personservice.persons.PersonService
 import com.hedvig.personservice.persons.model.PersonFlags
+import com.hedvig.personservice.persons.web.dtos.HasSignedBeforeRequest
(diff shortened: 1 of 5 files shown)


FredrikAreschoug (approved):

member-service/src/main/java/com/hedvig/personservice/persons/web/PersonController.kt
```diff
@@ -82,8 +88,8 @@ class PersonController @Autowired constructor(
     }
 
     @PostMapping("/has/signed")
-    fun hasSigned(@RequestBody request: HasPersonSignedBeforeRequest): ResponseEntity<Boolean> {
-        val hasSigned = personService.hasSigned(request.ssn, request.email)
+    fun hasSigned(@RequestBody request: HasSignedBeforeRequest): ResponseEntity<Boolean> {
```
johantjellden, line 91: Why did we put this logic on the person istead of on the member?

vonElfvin: Because you can sign with another member.

PR #440, opened 2020-12-10
HVG-645 simple sign event

FredrikAreschoug:
Start simple sign now signes member

Changes:
* added MemberSimpleSignEvent
* Extracted set birthdate from MemberEventListener and uses update ssn event instead
*  removed duplication on memberSigned
* added UpdateBirthDateCommand
* created correction endpoint


member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -117,6 +125,21 @@ void authAttempt(AuthenticationAttemptCommand command) {
     }
   }
 
+  private void getBirthdateFromSwedishSsnAndApplyEvent(Long memberId, String ssn) {
+    LocalDate birthDate = getBirthdateFromSwedishSsn(ssn);
+    if (birthDate != null) {
+      apply(new BirthDateUpdatedEvent(memberId, birthDate));
+    }
```
johantjellden, line 132: I honestly don't know if we need the applyChain but maybe we should investigate that another time. But cant this method take an optional applyChain then we can use it at both places.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -117,6 +125,21 @@ void authAttempt(AuthenticationAttemptCommand command) {
...
+  }
+
+  private LocalDate getBirthdateFromSwedishSsn(String ssn) {
+      if (ssn != null) {
+          final DateTimeFormatter dtf = DateTimeFormatter.ofPattern("yyyyMMdd");
+          return LocalDate.parse(ssn.substring(0, 8), dtf);
+      }
+      return null;
+  }
+
```
johantjellden, line 142: Would it make sense to move this function to some util package?

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -289,6 +314,12 @@ public void on(SignMemberFromUnderwriterCommand signMemberFromUnderwriterCommand
     apply(new MemberSignedWithoutBankId(signMemberFromUnderwriterCommand.getId(), signMemberFromUnderwriterCommand.getSsn()));
   }
 
+  @CommandHandler
+  public void on(MemberSimpleSignedCommand memberSimpleSignedCommand) {
+    apply(new MemberSimpleSignedEvent(memberSimpleSignedCommand.getId(), memberSimpleSignedCommand.getSsn(), memberSimpleSignedCommand.getReferenceId()));
+    apply(new SSNUpdatedEvent(this.id, memberSimpleSignedCommand.getSsn()));
```
johantjellden, line 320: Do we now which country the member is from? I mean if it is a Swedish, Danish or Norwegian ssn? If so then we could update the birthdate as well?

member-service/src/main/java/com/hedvig/memberservice/correction/CorrectSwedishSsnEventComponent.kt
```diff
@@ -0,0 +1,52 @@
...
+
+                if (birthDateUpdatedEvents.isEmpty()) {
+                    getBirthDateFromSwedishSsn(member.ssn)?.let {
+                        commandGateway.sendAndWait<Void>(UpdateBirthDateCommand(member.id, it))
+                        counter++
+                    }
+                }
+            }
+        }
+
```
johantjellden, line 40: Sneaky. Is the `member.ssn.length != 1` so that we do this only for swedish members?

johantjellden (commented):
Sweet! Minor comments.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
FredrikAreschoug: No we don't and Sweden is the only one of this that we actually know the birthdate from the ssn, in the other cases we could only do a good guess. Is it not better if we try to apply that another way?

member-service/src/main/java/com/hedvig/memberservice/correction/CorrectSwedishSsnEventComponent.kt
FredrikAreschoug: Yeah this will only work for now but yeah the danish ssn is 10 and the Norwegian one is 11 :slightly_smiling_face:

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
johantjellden: Yes I think that you are correct. And as a future enhancement should we have a property on the member that says which nationality it had?

member-service/src/main/java/com/hedvig/memberservice/correction/CorrectSwedishSsnEventComponent.kt
johantjellden: Yes but I guess that this is a one of :)

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.java
```diff
@@ -117,6 +123,20 @@ void authAttempt(AuthenticationAttemptCommand command) {
...
+            applyChain = applyChain.andThenApply(
+                () -> new BirthDateUpdatedEvent(this.id, birthDate)
+            );
+            return applyChain;
+        }
+      apply(new BirthDateUpdatedEvent(memberId, birthDate));
+    }
+    return applyChain;
+  }
+
```
johantjellden, line 139: Nice! If i remember correctly the aggregateId is the member id so I don't think we need to pass that along.

johantjellden: Or pass it along as we do with the other methods and use `memberId` on line 131

johantjellden (commented):
Two new minor comments

member-service/src/main/java/com/hedvig/memberservice/events/upcasters/OnboardingStartedWithSSNEventUpcaster.kt
```diff
@@ -0,0 +1,30 @@
+package com.hedvig.memberservice.events.upcasters
+
+import com.hedvig.memberservice.events.OnboardingStartedWithSSNEvent
+import org.axonframework.serialization.SimpleSerializedType
+import org.axonframework.serialization.upcasting.event.IntermediateEventRepresentation
+import org.axonframework.serialization.upcasting.event.SingleEventUpcaster
+
+class OnboardingStartedWithSSNEventUpcaster: SingleEventUpcaster() {
+    override fun canUpcast(intermediateRepresentation: IntermediateEventRepresentation): Boolean {
+        return intermediateRepresentation.type == targetType
```
johantjellden, line 10: Do we need to specify something to get this to only be run when there is no version information on the event?

member-service/src/main/java/com/hedvig/memberservice/events/upcasters/SSNUpdatedEventUpcaster.kt
```diff
@@ -0,0 +1,29 @@
+package com.hedvig.memberservice.events.upcasters
+
+import com.hedvig.memberservice.events.SSNUpdatedEvent
+import org.axonframework.serialization.SimpleSerializedType
+import org.axonframework.serialization.upcasting.event.IntermediateEventRepresentation
+import org.axonframework.serialization.upcasting.event.SingleEventUpcaster
+
+class SSNUpdatedEventUpcaster: SingleEventUpcaster() {
+    override fun canUpcast(intermediateRepresentation: IntermediateEventRepresentation): Boolean {
+        return intermediateRepresentation.type == targetType
```
johantjellden, line 10: Same question as with the other upcaster :)

johantjellden (commented):
Looks good :) Two questions ;)

member-service/src/main/java/com/hedvig/memberservice/events/upcasters/OnboardingStartedWithSSNEventUpcaster.kt
FredrikAreschoug: In the targetType line 26, we have revision number `null` which was the revision before `1.0`

member-service/src/main/java/com/hedvig/memberservice/events/upcasters/SSNUpdatedEventUpcaster.kt
FredrikAreschoug: Same answer here :slightly_smiling_face:

member-service/src/main/java/com/hedvig/memberservice/events/upcasters/OnboardingStartedWithSSNEventUpcaster.kt
johantjellden: 🥇

johantjellden (approved):
Sweet 🚀

PR #445, opened 2020-12-14
Cleanup customer.io from Norwegian, Danish and Rapio members (also SimpleSign as well)

vonElfvin:


FredrikAreschoug (approved):
Looks good to me

PR #446, opened 2020-12-14
Rename processing group to something more correct and that we backfill as a bonus

vonElfvin:
I'll merge this tomorrow morning if we see that the fix works


FredrikAreschoug (approved):
:+1:

PR #456, opened 2021-01-11
Store and use strategy

FredrikAreschoug:
Fixes the problem of not getting status form the correct place

### DIFF
--- a/member-service/src/main/java/com/hedvig/memberservice/entities/UnderwriterSignSessionEntity.kt
+++ b/member-service/src/main/java/com/hedvig/memberservice/entities/UnderwriterSignSessionEntity.kt
@@ -1,7 +1,10 @@
 package com.hedvig.memberservice.entities
 
+import com.hedvig.memberservice.services.signing.underwriter.strategy.SignStrategy
 import java.util.*
 import javax.persistence.Entity
+import javax.persistence.EnumType
+import javax.persistence.Enumerated
 import javax.persistence.GeneratedValue
 import javax.persistence.GenerationType
 import javax.persistence.Id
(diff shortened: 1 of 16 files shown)


member-service/src/main/java/com/hedvig/memberservice/query/UnderwriterSignSessionRepository.kt
```diff
@@ -25,10 +27,13 @@ interface UnderwriterSignSessionRepository: JpaRepository<UnderwriterSignSession
         UPDATE UnderwriterSignSessionEntity session set session.underwriterSignSessionReference = ?1 where session.internalId = ?2
     """)
     fun updateUnderwriterSignSessionReference(underwriterSignSessionReference: UUID?, internalId: Long)
+
+    @Query
+    fun findTopByMemberIdOrderByInternalIdDesc(memberId: Long): UnderwriterSignSessionEntity?
```
meletios, line 32: Sweet

member-service/src/main/java/com/hedvig/memberservice/entities/UnderwriterSignSessionEntity.kt
```diff
@@ -12,10 +15,13 @@ import javax.persistence.UniqueConstraint
 @Table(uniqueConstraints=[UniqueConstraint(columnNames = ["underwriterSignSessionReference", "signReference"])])
 class UnderwriterSignSessionEntity(
     val underwriterSignSessionReference: UUID,
-    val signReference: UUID
+    val signReference: UUID,
+    val memberId: Long?,
```
meletios, line 19: Can we have the memberId as non-nullable?

meletios (approved):
Sweet, i have one small question though

member-service/src/main/java/com/hedvig/memberservice/entities/UnderwriterSignSessionEntity.kt
FredrikAreschoug: No we have already a lot of sessions stored so it will be null in those cases

PR #457, opened 2021-01-15
[HVG-268] Trustpilot review invitation

bystam:
# Jira Issue: [HVG-268]

## What?
This implements an integration with Trustpilots API, where we authenticate using Oauth2 and with the credentials of a specific business account and then leverage that access token to create review links for our service.

These links are created and updated whenever members finalise the onboarding or if they change their name and/or email at a later time. That's because the links are pre-populated with this information in the trustpilot systems.

The created links are send as an attribute on CustomerIO entries through notification-service.

## Why?
This is part of project Bolt to increase our trustpilot score.

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [x] Boyscouted

[HVG-268]: https://hedvig.atlassian.net/browse/HVG-268


member-service/src/main/java/com/hedvig/memberservice/services/customerio/CustomerIOEventListener.kt
```diff
@@ -0,0 +1,118 @@
...
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.context.annotation.Profile
+import org.springframework.stereotype.Component
+import java.util.Locale
+import java.util.Objects
+
+@Component
+@Profile("customer.io")
+@ProcessingGroup("SegmentProcessorGroup")
+class CustomerIOEventListener @Autowired constructor(
```
bystam, line 23: We prefixed it with `CustomIO` as we thought simply `EventListener` was a bit brief, even if it was part of the package.

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotOauth2Configuration.kt
```diff
@@ -0,0 +1,18 @@
...
+    val basePath: String,
+    @Value("\${trustpilot.oauth.apikey}")
+    val apikey: String,
+    @Value("\${trustpilot.oauth.secret}")
+    val secret: String,
+    @Value("\${trustpilot.oauth.username}")
+    val username: String,
+    @Value("\${trustpilot.oauth.password}")
+    val password: String
+)
```
bystam, line 18: Is this how you do config? 😅

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotFeignConfig.kt
```diff
@@ -0,0 +1,99 @@
...
+    private var accessTokenExpiryTime: Instant = Instant.now()
+    private var refreshTokenExpiryTime: Instant = Instant.now()
+
+    private lateinit var accessToken: String
+    private lateinit var refreshToken: String
+
+    override fun apply(template: RequestTemplate) {
+        if (Instant.now().isAfter(accessTokenExpiryTime)) {
+            authorize()
+        }
```
bystam, line 36: This interceptor is stateful, and runs in a multithreaded environment. The existing open source one (that we ultimately couldn't use) didn't contain any synchronization, so we ended up not adding any here either.

But it's not obvious to me that that's the right choice. If we wish to protect the reads and write of the properties in here, we could easily wrap this method in a `synchronized(this) { ... }` which should do the trick.

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotFeignConfig.kt
```diff
@@ -0,0 +1,99 @@
...
+class TrustpilotFeignConfig {
+
+    @Bean
+    fun oauth2RequestInterceptor(template: RestTemplate, config: TrustpilotOauth2Configuration): RequestInterceptor {
+        return TrustpilotOauth2Interceptor(template, config)
+    }
+}
+
+private class TrustpilotOauth2Interceptor(
+    val template: RestTemplate,
```
bystam, line 23: We chose to use `RestTemplate` instead of a nested `FeignClient` here - because the nesting felt a bit much, and the newer spring `WebClient` wasn't currently part of the existing dependencies.

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotClient.kt
```diff
@@ -0,0 +1,30 @@
...
+    val referenceId: String,
+    val email: String,
+    val name: String,
+    val locale: String
+)
+
+data class TrustpilotReviewLinkResponseDto(
+    val id: String,
+    val url: String
+)
```
meletios, line 30: Maybe separate those to different files?

meletios (commented):
Looks good overall. A general question though. Whats your opinion on having classes in separate files? I think the files would look a bit smaller and easier to understand.

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotClient.kt
bystam: My preference usually is that if there are a few types that are very small, and they go together, I personally thinks it helps readability when they are close together, as you can visually see them all at once instead of having to look around in multiple files.

But I am happy to abide by conventions you usually use here :) I don't have a strong opinion, but instead usually just make decisions about separate files on not on a case-by-case basis.

vonElfvin: I prefer different files as well, they can stay in the same/close folder though so they stay together

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotFeignConfig.kt
```diff
@@ -0,0 +1,99 @@
...
+
+class TrustpilotFeignConfig {
+
+    @Bean
+    fun oauth2RequestInterceptor(template: RestTemplate, config: TrustpilotOauth2Configuration): RequestInterceptor {
+        return TrustpilotOauth2Interceptor(template, config)
+    }
+}
+
+private class TrustpilotOauth2Interceptor(
```
vonElfvin, line 22: Separate file pls, also maybe this one can be generic

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotOauth2Configuration.kt
vonElfvin: I'd say no, but Meletis seems to agree? I'd just inject them when/where you need them instead. And remove this @Configuration entirely since it has no beans

member-service/src/main/java/com/hedvig/memberservice/services/customerio/CustomerIOEventListener.kt
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+        val traits = mutableMapOf(
+            "first_name" to evt.firstName,
+            "last_name" to evt.lastName
+        )
+
+        if (trustPilotLinkResponse != null) {
+            traits["trustpilotLink"] = trustPilotLinkResponse.url
+        }
```
vonElfvin, line 52: The trustpilot part could be added via a helper function that is reused by both

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotClient.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import org.springframework.cloud.openfeign.FeignClient
+import org.springframework.web.bind.annotation.*
+
+@FeignClient(
+    name = "trustpilotClient",
+    url = "https://invitations-api.trustpilot.com/v1/private",
+    configuration = [TrustpilotFeignConfig::class]
+)
+interface TrustpilotClient {
```
vonElfvin, line 11: I personally dislikt having this inside member-service if we want to reuse it somethwere else. I'd put this in lookup-service. But I guess we can move it at that point if it comes.

member-service/src/main/java/com/hedvig/memberservice/services/trustpilot/TrustpilotReviewService.kt
```diff
@@ -0,0 +1,17 @@
...
+interface TrustpilotReviewService {
+    fun generateTrustpilotReviewInvitation(
+        memberId: Long,
+        email: String,
+        name: String,
+        locale: Locale?
+    ): TrustpilotReviewInvitation?
+}
+
+data class TrustpilotReviewInvitation(
```
vonElfvin, line 14: Separate file pls

vonElfvin (commented):
Really neat! 👍

member-service/src/main/java/com/hedvig/memberservice/services/customerio/CustomerIOEventListener.kt
vonElfvin: Also I think it should be `trustpilot_link` since we usually use snakecase for attributes.

vonElfvin: And we should probably add `trustpilot_link_id` in case they change/update the basepath on trustpilot's end

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotClient.kt
bystam: Moving it to lookup-service will arguably increase the complexity of the system, though, by drawing a new arrow between the services. My experience at Kry is that the more frequently one service talks to another one, the harder it will be to run things locally, and that that difficulty will grow exponentially.

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotClient.kt
bystam: Alright! I will update it

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotFeignConfig.kt
bystam: Will do. I put it here because it was able to be `private` but happy to move it.

Is it worthwhile making it generic before it's actually reused? Feels a bit like a premature abstraction to me.

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotOauth2Configuration.kt
bystam: That makes sense - maybe I was thinking of `@ConfigurationProperties` (https://www.baeldung.com/configuration-properties-in-spring-boot) - which I am not sure we're using here in member service?

member-service/src/main/java/com/hedvig/memberservice/services/customerio/CustomerIOEventListener.kt
bystam: Will fix

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotOauth2Configuration.kt
bystam: I had a bunch of trouble modelling this, because the `FeignClient` config object didn't want to be a `@Configuration` really, and I just re-designed how these values were injected like 5 times just to have it work.

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotClient.kt
johantjellden: I also think that coupling services together complicates things a lot. So i would also vote for keeping this inside of member-service for now.

member-service/src/test/java/com/hedvig/memberservice/services/customerio/CustomerIOEventListenerTest.kt
```diff
@@ -0,0 +1,164 @@
...
+
+        val sut = CustomerIOEventListener(notificationService, memberRepository, trustpilotReviewService)
+
+        every {
+            trustpilotReviewService.generateTrustpilotReviewInvitation(any(), any(), any(), any())
+        } returns TrustpilotReviewInvitation("id", "url")
+
+        sut.on(EmailUpdatedEvent(1337L, "omse@lkj.com"))
+    }
+}
```
johantjellden, line 164: I like to use the `import org.junit.rules.ExpectedException` rule to test for exceptions. I would look something like this:
```
    @get:Rule
    val thrown = ExpectedException.none()
    @Test
    fun `no memberentity found fails`() {

        every { memberRepository.findById(any()) } returns Optional.empty()

        val sut = CustomerIOEventListener(notificationService, memberRepository, trustpilotReviewService)

        thrown.expect(IllegalStateException::class.java)
        sut.on(EmailUpdatedEvent(1337L, "omse@lkj.com"))
    }
```

Anyway I think that the mock setup should be removed for clarity :)

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotOauth2Interceptor.kt
```diff
@@ -0,0 +1,97 @@
...
+        if (Instant.now().isAfter(accessTokenExpiryTime)) {
+            authorize()
+        }
+
+        template.header("Authorization", "Bearer $accessToken")
+    }
+
+    private fun authorize() {
+        if (Instant.now().isBefore(refreshTokenExpiryTime)) {
+            refreshAccessToken(refreshToken)
```
johantjellden, line 41: I saw your comment on the old interceptor about state and threads. Perhaps it is a good idea to synchronize here in order to avoid any confusing behavior in the future?

bystam: Great that you kept track of that comment, I didn't realise it disappeared 😅 

Yes, I agree. We definitely won't be suffering any performance issues from the synchronisation, so I agree - let's add it :)

member-service/src/test/java/com/hedvig/memberservice/services/customerio/CustomerIOEventListenerTest.kt
bystam: Thanks for the tips! I will update this

member-service/src/main/java/com/hedvig/memberservice/services/customerio/CustomerIOEventListener.kt
```diff
@@ -0,0 +1,119 @@
...
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.context.annotation.Profile
+import org.springframework.stereotype.Component
+import java.util.Locale
+import java.util.Objects
+
+@Component
+@Profile("customer.io")
+@ProcessingGroup("SegmentProcessorGroup")
+class CustomerIOEventListener @Autowired constructor(
```
FredrikAreschoug, line 22: `@Autowired constructor` is not needed but it's always added when converting from java
```suggestion
class CustomerIOEventListener(
```

member-service/src/test/java/com/hedvig/memberservice/external/trustpilot/TrustpilotOauth2InterceptorTest.kt
```diff
@@ -0,0 +1,112 @@
...
+
+import com.hedvig.memberservice.external.trustpilot.TrustpilotOauth2Interceptor.TokenResponse
+import feign.RequestTemplate
+import io.mockk.MockKAnnotations
+import io.mockk.every
+import io.mockk.impl.annotations.MockK
+import io.mockk.slot
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Before
+import org.junit.Test
```
FredrikAreschoug, line 11: Personal opinion, when creating new tests I like using junit 5 and it looks like they will run if you would change this to and use `BeforeEach` annotation instead of `Before` :slightly_smiling_face:  But as mentioned Personal opinion
```suggestion
import org.junit.jupiter.api.BeforeEach
import org.junit.jupiter.api.Test
```

FredrikAreschoug (approved):
Clean! Looks good to me, added two minor comments, do as you please with them :slightly_smiling_face:

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotFeignConfig.kt
FredrikAreschoug: In my opinion I don't think we have to expose this class to the rest of the application. I would prefer doing it like this more

member-service/src/test/java/com/hedvig/memberservice/external/trustpilot/TrustpilotOauth2InterceptorTest.kt
bystam: Aaah - I was used to `@BeforeEach` too - but I didn't get that to work. I realise now that it didn't because I was still using the Junit4 version of `@Test` of course

member-service/src/main/java/com/hedvig/memberservice/external/trustpilot/TrustpilotFeignConfig.kt
bystam: Uh oh ^^ who has the stronger opinion here? :P

vonElfvin: I prefer clear files with a solid structure where you don't mix in classes with files of others. And I don't see the risk of exposing it being big at all. But you decide, you know more about this than me

johantjellden (dismissed):
💯

member-service/src/main/java/com/hedvig/memberservice/services/customerio/CustomerIOTrustpilotEventListener.kt
```diff
@@ -0,0 +1,97 @@
...
+@Profile("customer.io")
+@ProcessingGroup("CustomerIOTrustpilotBackfill")
+class CustomerIOTrustpilotEventListener(
+    private val notificationService: NotificationService,
+    private val memberRepository: MemberRepository,
+    private val trustpilotReviewService: TrustpilotReviewService
+) {
+
+    @EventHandler
+    fun on(evt: NameUpdatedEvent) {
```
vonElfvin, line 28:
```suggestion
    fun on(event: NameUpdatedEvent) {
```

member-service/src/main/java/com/hedvig/memberservice/services/customerio/CustomerIOTrustpilotEventListener.kt
```diff
@@ -0,0 +1,97 @@
...
+        val member = memberRepository.findById(evt.memberId).orElseThrow {
+            IllegalStateException("No member found for ${evt.memberId}")
+        }
```
vonElfvin, line 31: A bit more clear logging IMO
```suggestion
        val member = memberRepository.findById(evt.memberId).orElseThrow {
            IllegalStateException("No member found when generating Trustpilot link (memberId=${event.memberId})")
        }
```

member-service/src/main/java/com/hedvig/memberservice/services/customerio/CustomerIOTrustpilotEventListener.kt
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    private fun getLocaleForMember(member: MemberEntity): Locale? {
+        return member.pickedLocale?.locale
+            ?: LocaleResolver.resolveNullableLocale(member.acceptLanguage)
+    }
+
+    private fun createTrustpilotReviewAttributes(
+        memberId: Long, locale: Locale?, email: String?, firstName: String?, lastName: String?
+    ): Map<String, String> {
```
vonElfvin, line 72:
```suggestion
    private fun createTrustpilotReviewAttributes(
        memberId: Long, 
        locale: Locale?,
        email: String?, 
        firstName: String?, 
        lastName: String?
    ): Map<String, String> {
```

member-service/src/main/java/com/hedvig/memberservice/services/customerio/CustomerIOTrustpilotEventListener.kt
```diff
@@ -0,0 +1,97 @@
...
+import org.axonframework.config.ProcessingGroup
+import org.axonframework.eventhandling.EventHandler
+import org.springframework.context.annotation.Profile
+import org.springframework.stereotype.Component
+import java.util.Locale
+import java.util.Objects
+
+@Component
+@Profile("customer.io")
+@ProcessingGroup("CustomerIOTrustpilotBackfill")
```
vonElfvin, line 20: Seems to me like you won't kill off this ProcessingGroup? May as well remove the word "Backfill"

The word "Backfill" to me sounds like it will be removed after being ran. 
```suggestion
@ProcessingGroup("CustomerIOTrustpilot")
```

member-service/src/main/java/com/hedvig/memberservice/services/trustpilot/TrustpilotReviewServiceImpl.kt
```diff
@@ -0,0 +1,37 @@
...
+        name: String,
+        locale: Locale?
+    ): TrustpilotReviewInvitation? {
+        val locale = locale ?: LocaleResolver.DEFAULT_LOCALE
+        return try {
+            val response = trustpilotClient.createReviewLink(
+                businessUnitId,
+                TrustpilotReviewLinkRequestDto(
+                    memberId.toString(), email, name, "${locale.language}-${locale.country}"
+                )
```
vonElfvin, line 28: I prefer having the name on there as well for extra readability
```suggestion
                TrustpilotReviewLinkRequestDto(
                    memberId.toString(), 
                    email, 
                    name, 
                    "${locale.language}-${locale.country}"
                )
```

vonElfvin (dismissed):
Really good, this will work wonders! Just some code styling/naming to consider if you want

member-service/src/main/java/com/hedvig/memberservice/services/customerio/CustomerIOTrustpilotEventListener.kt
bystam: Right - I figured the stack trace would be enough but even clearer messages are always good

member-service/src/main/java/com/hedvig/memberservice/services/customerio/CustomerIOTrustpilotEventListener.kt
```diff
@@ -0,0 +1,97 @@
...
+
+        if (attributes.isNotEmpty()) {
+            sendWithSleep(attributes, Objects.toString(evt.memberId))
+        }
+    }
+
+    @EventHandler
+    fun on(evt: EmailUpdatedEvent) {
+        val member = memberRepository.findById(evt.memberId).orElseThrow {
+            IllegalStateException("No member found for ${evt.memberId}")
```
meletios, line 49: Detailed message here as well

vonElfvin (approved):

PR #458, opened 2021-01-18
split out trustpilot part of customerio to make sure it can be backfilled correctly

bystam:
# Jira Issue: [HVG-268]

## What?
The trustpilot part of the customer.io event listener needed backfilling that the existing code didn't need, therefore we split it out into a new event listener to have it run on a separate processing group.

We also added an `@Order` annotation to the existing `MemberEventListener` to make sure that **that** one always runs before the other ones, fearing we would otherwise introduce a race condition. Our event listener depends on the values that are updated by these events.

## Why?
- 

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [x] Codescouted

[HVG-268]: https://hedvig.atlassian.net/browse/HVG-268


member-service/src/main/java/com/hedvig/memberservice/query/MemberEventListener.java
```diff
@@ -12,15 +12,16 @@
...
 import org.springframework.stereotype.Component;
 
 import java.time.Instant;
-import java.time.LocalDate;
-import java.time.format.DateTimeFormatter;
 import java.util.concurrent.CompletableFuture;
 
 @Component
 @Slf4j
+@Order(Ordered.HIGHEST_PRECEDENCE)
```
bystam, line 24: We were unsure whether this was correct or not. Reading the documentation, it kind of feels like it's **not** the correct approach? Rather feels like maybe this one should be a "subscribing event processor" instead of a tracking one? 🤔 

https://docs.axoniq.io/reference-guide/v/4.0/configuring-infrastructure-components/event-processing/event-processors

member-service/src/main/java/com/hedvig/memberservice/backfill/BackfillMemberIdentifiedEventNorwayListener.kt
```diff
@@ -1,74 +0,0 @@
...
-import java.time.Instant
-import java.time.LocalDateTime
-import java.time.ZoneOffset
-
-/**
- * This is a one of and will be removed once its is used once
- */
-@Component
-@ProcessingGroup("BackfillMemberIdentifiedEventNorway")
-class BackfillMemberIdentifiedEventNorwayListener(
```
johantjellden, line 24: Should we remove this from our `AxonConfig.java`?

member-service/src/main/java/com/hedvig/memberservice/query/MemberEventListener.java
johantjellden: Hmm, I think that order only matters for EventListeners that are in the same processing group. But I could be wrong, I will check the documentation.

bystam: Yes - that's exactly right. So I think we might be at a place where none of us were entirely certain how to model this.

Our new EventListener listens to two events, but also depends on the `MemberRepository` - or more specifically depends on the changes of those same events into that repository. As I understand it, all event listeners run on threads managed by their processing group (if not otherwise specified), which would mean that there's a race condition between our listener and the `MemberEventListener` one.

I have ideas on how I would **want** to model this (I am used to reactive stream APIs 😅) but I don't know how to translate that into Axon (or if I even need to? Are `java.concurrency` types okay to use?).

bystam: We had some discussions between me and @AlexSJ1990 whether this was an atually decent case for using a Saga, but she also mentioned that usages of a Saga had been something you were currently re-evaluating a bit internally :)

member-service/src/main/java/com/hedvig/memberservice/backfill/BackfillMemberIdentifiedEventNorwayListener.kt
bystam: I'm confused by this diff - it must have ended up here because of a merge with master

PR #462, opened 2021-01-22
Fix/simple sign login issue

FredrikAreschoug:
## What?
- Fix for simple sign login issue

## Why?
- Because you cant authenticate if you signed with simple sign 

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
```diff
@@ -31,19 +32,44 @@ class ZignSecBankIdService(
...
-        ZignSecBankIdAuthenticationRequest(
-            memberId.toString(),
-            request.personalNumber,
-            resolveTwoLetterLanguageFromMember(memberId),
-            authenticationSuccessUrl,
-            authenticationFailUrl,
-            zignSecAuthenticationMarket.getAuthenticationMethod()
-        )
-    )
+        zignSecAuthenticationMarket: ZignSecAuthenticationMarket) = when (zignSecAuthenticationMarket) {
```
bystam, line 42: Personal preference, not sure how you feel about it here at Hedvig, but usually if the body is rather compex with `when` branching and stuff, I think it's nice to spell out the return type explicitly to make it more clear what this actually returns. Especially since the return type in some branches is determined by a nested method call like `zignSecAuthentication.auth`

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
```diff
@@ -31,19 +32,44 @@ class ZignSecBankIdService(
...
+                        memberId.toString(),
+                        it,
+                        resolveTwoLetterLanguageFromMember(memberId),
+                        authenticationSuccessUrl,
+                        authenticationFailUrl,
+                        zignSecAuthenticationMarket.getAuthenticationMethod()
+                    )
+                )
+            } ?: StartZignSecAuthenticationResult.Success(
+                UUID.randomUUID(), // should not be used
```
bystam, line 56: If this should be used, would it make sense to make it nullable instead of something random like this?

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
bystam: If nothing else it will help compile times ^^ as I'm pretty sure type inference usually is one of the most expensive parts of compiling.

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
bystam: Maybe I'm totally misunderstanding what this value actually is.

vonElfvin (approved):
Looks good to me. After discussion I'd say this will be solid. Added some  optionalcode touches

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
```diff
@@ -93,16 +121,25 @@ class ZignSecBankIdService(
...
+        Locale("nb").language -> baseUrl + NORWEGIAN_BANK_ID_NORWEGIAN_LOGIN_PATH
+        else -> baseUrl + NORWEGIAN_BANK_ID_ENGLISH_LOGIN_PATH
+    }
+
+    private fun resolveTwoLetterLanguageFromMember(memberId: Long, acceptLanguage: String?) = when (resolveLocaleFromMember(memberId, acceptLanguage)?.language) {
+        Locale("sv").language -> "SV"
+        Locale("en").language -> "EN"
+        Locale("nb").language -> "NO"
+        Locale("da").language -> "DA"
+        else -> "EN"
```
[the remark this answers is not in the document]
FredrikAreschoug: this is only for zignsec so I don't agree

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
```diff
@@ -26,24 +29,47 @@ class ZignSecBankIdService(
...
-            memberId.toString(),
-            request.personalNumber,
-            resolveTwoLetterLanguageFromMember(memberId),
-            authenticationSuccessUrl,
-            authenticationFailUrl,
-            zignSecAuthenticationMarket.getAuthenticationMethod()
+        zignSecAuthenticationMarket: ZignSecAuthenticationMarket,
+        acceptLanguage: String?
+    ): StartZignSecAuthenticationResult {
+        if (zignSecAuthenticationMarket == ZignSecAuthenticationMarket.NORWAY &&
```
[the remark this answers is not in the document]
FredrikAreschoug: I agree that that would be better if `request.personalNumber == null` would always become a static redirect but throwing in denmark would not be the ideal case.

PR #464, opened 2021-01-27
take care of d number and norwegian bankId bug

FredrikAreschoug:
# Jira Issue: [HVG-848] 

## What?
- Take care of d number and Norwegian BankId bug

## Why?
- We want everyone to be able to sign

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted

[HVG-848]: https://hedvig.atlassian.net/browse/HVG-848

### DIFF
--- a/zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
+++ b/zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
@@ -314,7 +314,7 @@ class ZignSecSessionServiceImpl(
     }
 
     /**
-     * On 26:th of january 2021 we found a bug where norwegian bank id returns the birthdate of by one day
+     * On 26:th of january 2021 we found a bug where norwegian bank id returns the birthdate off by one day
      *
      * Note: To this date this has never changed month so for now this is not something we want to include in this fix
      */
--- a/zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
+++ b/zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.hedvig.external.authentication.dto.ZignSecAuthenticationResult
 import com.hedvig.external.authentication.dto.ZignSecSignResult
 import com.hedvig.external.authentication.dto.StartZignSecAuthenticationResult
+import com.hedvig.external.authentication.dto.ZignSecAuthenticationMethod
 import com.hedvig.external.authentication.dto.ZignSecAuthenticationResponseError
 import com.hedvig.external.authentication.dto.ZignSecBankIdAuthenticationRequest
 import com.hedvig.external.authentication.dto.ZignSecBankIdProgressStatus
@@ -18,6 +19,7 @@ import com.hedvig.external.zignSec.repository.entitys.ZignSecSession
 import com.hedvig.external.zignSec.repository.entitys.ZignSecSignEntity
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Service
+import java.text.DecimalFormat
 
 @Service
 class ZignSecSessionServiceImpl(
@@ -191,11 +193,9 @@ class ZignSecSessionServiceImpl(
                     }
                     ZignSecBankIdProgressStatus.COMPLETED -> {
                         metrics.authSessionSuccess(session.authenticationMethod, session.requestType)
-                        //TODO: re add this when everything is in order with zign sec and person number also un ignore the test in ZignSecSessionServiceImplTest
-                        //NOTE: this is also used in denmark so make sure both work on changing
-                        //check that personal number is matching when signing
-//                        if (notification.identity!!.personalNumber!! != session.requestPersonalNumber!!) {
-                        if (notification.identity!!.dateOfBirth!!.dayMonthAndTwoDigitYearFromDateOfBirth() != session.requestPersonalNumber!!.dayMonthAndTwoDigitYearFromNorwegianOrDanishSsn()) {
+                        if (validatePersonNumberAgainstDateOfBirth(session.requestPersonalNumber!!, notification.identity!!.dateOfBirth!!, session.authenticationMethod)) {
+                            memberSigned(session, jsonRequest, notification.identity.firstName, notification.identity.lastName)
+                        } else {
                             session.status = ZignSecBankIdProgressStatus.FAILED
                             authenticationEventPublisher.publishSignEvent(
                                 ZignSecSignResult.Failed(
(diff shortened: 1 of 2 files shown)


zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
```diff
@@ -293,23 +300,73 @@ class ZignSecSessionServiceImpl(
...
 
-fun String.dayMonthAndTwoDigitYearFromNorwegianOrDanishSsn(): Triple<String, String, String> {
-    val trimmedInput = this.trim().replace("-", "").replace(" ", "")
-    val day = trimmedInput.substring(0, 2)
-    val month = trimmedInput.substring(2, 4)
-    val twoDigitYear = trimmedInput.substring(4, 6)
-    return Triple(day, month, twoDigitYear)
-}
+    /**
+     * On 26:th of january 2021 we found a bug where norwegian bank id returns the birthdate of by one day
```
Michael-Duivestein, line 317:
```suggestion
     * On 26:th of january 2021 we found a bug where norwegian bank id returns the birthdate off by one day
```

zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
```diff
@@ -293,23 +300,73 @@ class ZignSecSessionServiceImpl(
...
-    companion object {
-        private val logger = LoggerFactory.getLogger(ZignSecServiceImpl::class.java)
+    fun validatePersonNumberAgainstDateOfBirth(personNumber: String, dateOfBirth: String, method: ZignSecAuthenticationMethod): Boolean {
+        if (personNumber.dayMonthAndTwoDigitYearFromNorwegianOrDanishSsn() == dateOfBirth.dayMonthAndTwoDigitYearFromDateOfBirth()) {
+            return true
+        }
+
+        return when (method) {
+            ZignSecAuthenticationMethod.NORWAY_WEB_OR_MOBILE ->
+                validatePersonNumberAgainstDateOfBirthWithKnownNorwegianBankIdBug(personNumber, dateOfBirth) ||
```
vonElfvin, line 310: Unsure where to put the suggestion, but working with LocalDates would make this a lot easier. Especially since "plusDays" and "minusDays" will take into account if the day is the last of the month.
```suggestion
        val birthDate = LocalDate....from(dateOfBirth)
        val ssnBirthDate = LocalDate....from(personNumber)
        
        return when (method) {
            ZignSecAuthenticationMethod.NORWAY_WEB_OR_MOBILE ->
                !birthDate.isBefore(ssnBirthDate.minusDays(4) && !birthDate.isAfter(ssnBirthDate.plusDays(4))
```

zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
```diff
@@ -293,23 +300,73 @@ class ZignSecSessionServiceImpl(
...
+    /**
+     * On 26:th of january 2021 we found a bug where norwegian bank id returns the birthdate of by one day
+     *
+     * Note: To this date this has never changed month so for now this is not something we want to include in this fix
+     */
+    fun validatePersonNumberAgainstDateOfBirthWithKnownNorwegianBankIdBug(personNumber: String, dateOfBirth: String): Boolean {
+        val personNumberTriple = personNumber.dayMonthAndTwoDigitYearFromNorwegianOrDanishSsn()
+        val dayMonthAndTwoDigitYearTriple = dateOfBirth.dayMonthAndTwoDigitYearFromDateOfBirth()
+
+        val birthdayIncreasedByOne = DATE_FORMAT.format(dayMonthAndTwoDigitYearTriple.first.toInt() + 1)
```
vonElfvin, line 325: This will break if the birthDAte is e.g. 31st of January

vonElfvin (commented):
I think working with LocalDates would make this more robust

zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
bystam: +1. The `java.time` package is amazing and the `Local*` types are perfect for doing date algebra like this.

zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
```diff
@@ -293,23 +300,73 @@ class ZignSecSessionServiceImpl(
...
+        val twoDigitYear = trimmedInput.substring(4, 6)
+        return Triple(day, month, twoDigitYear)
+    }
+
+    //Format `yyyy-mm-dd`
+    fun String.dayMonthAndTwoDigitYearFromDateOfBirth(): Triple<String, String, String> {
+        val twoDigitYear = this.substring(2, 4)
+        val month = this.substring(5, 7)
+        val day = this.substring(8, 10)
+        return Triple(day, month, twoDigitYear)
```
bystam, line 365: `LocalDate.parse(...)` should create a similar structure as this but with a nicer API

zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
FredrikAreschoug: I chose to not use Dates because of 2 reasons 1. We don't actually have the year in the ssn 2. It will not work with when it is actually not a date. Like 33:th of January.

zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
FredrikAreschoug: I did not test the 31st of January but I did test 29th and that should produce the same error as you referring to right please have a look at `validateNorwegianDNumberAgainstDateOfBirth`

zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
FredrikAreschoug: true but matching it with the ssn that doesn't have a four-digit representation of the year would be harder.

zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
FredrikAreschoug: True it maybe its better to do that assumption I can change this method to do use local date But I don't agree on using that for d numbers because when it comes to the ssn it is not really a date that we are talking about.

zign-sec/src/main/java/com/hedvig/external/zignSec/ZignSecSessionServiceImpl.kt
bystam: Oh you're right I didn't particularly realise it was a personal ID number and not an ISO date. My bad!

vonElfvin (approved):

PR #466, opened 2021-01-29
Place the trustpilot integration behind a feature flag

bystam:
# Jira Issue: -

## What?
After an application-development.yml update I lost my trustpilot credentials, and got annoyed with myself immediately. Therefore I decided to jump on the codescouting train and **conditinally** load the trustpilot event listener bean, to make it only actually run if `trustpilot.customerio-review-links-enabled` is `true`.

This is an incremental step towards fewer dependencies on updated `application-development.yml` files.

## Why?
Developers shouldn't need to juggle trustpilot credentials when doing local development, because arguably the trustpilot integration shouldn't run at all when developing locally.

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted


AlexSJ1990 (commented):
think this looks really good, but couldn't we achieve the same thing by using a profile e.g.
`@Profile("trustpilot")` on the TrustPilotServiceImpl

bystam:
> think this looks really good, but couldn't we achieve the same thing by using a profile e.g.
> `@Profile("trustpilot")` on the TrustPilotServiceImpl

I based this on an article that Johan shared that proposed not to over-use Profiles but instead use specific feature flags.

https://reflectoring.io/spring-boot-profiles/

AlexSJ1990 (approved):
Looks really nice!

PR #467, opened 2021-02-01
Add the correct encoder and decoder for the feign client

bystam:
# Jira Issue: -

## What?
The existing manually built feign client was missing the encoder/decoder that it should have been given from the spring environment. This will fix that.

## Why?
-

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted

### DIFF
--- a/member-service/src/main/java/com/hedvig/memberservice/services/trustpilot/TrustpilotConfig.kt
+++ b/member-service/src/main/java/com/hedvig/memberservice/services/trustpilot/TrustpilotConfig.kt
@@ -9,6 +9,7 @@ import feign.Feign
 import feign.codec.Decoder
 import feign.codec.Encoder
 import org.springframework.beans.factory.annotation.Value
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.cloud.openfeign.FeignClientsConfiguration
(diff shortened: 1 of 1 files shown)


member-service/src/main/java/com/hedvig/memberservice/services/trustpilot/TrustpilotConfig.kt
```diff
@@ -4,33 +4,43 @@ import com.hedvig.integration.notificationService.NotificationService
...
     ): CustomerIOTrustpilotEventListener {
         return CustomerIOTrustpilotEventListener(
             notificationService = notificationService,
             memberRepository =  memberRepository,
             trustpilotReviewService = TrustpilotReviewServiceImpl(
                 trustpilotClient = Feign.builder()
-                    .contract(SpringMvcContract()) // enables spring annotations like @PostMapping
+                    .encoder(encoder)
+                    .decoder(decoder)
+                    .contract(contract)
```
bystam, line 44: The `Feign.builder()` didn't receive the default beans for encoder, decoder and contract, but instead had some bare-bones deafult ones. This adds the spring ones which makes all of the spring default behaviours work.

johantjellden (approved):
👍

PR #468, opened 2021-02-03
simple sign attach token in static redirect to keep it around

bystam:
# Jira Issue: [HVG-862]

## What?
Attach the Authorization header as a URL parameter sent back when starting the static simple sign.

## Why?
The new static redirect we have needs to keep track of the token in order to be able to attach it later. Therefore we keep it around in a URL parameter.

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [x] Codescouted

[HVG-862]: https://hedvig.atlassian.net/browse/HVG-862

### DIFF
--- a/member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
+++ b/member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
@@ -12,11 +12,9 @@ import com.hedvig.memberservice.query.MemberRepository
 import com.hedvig.memberservice.query.SignedMemberRepository
 import com.hedvig.memberservice.services.redispublisher.AuthSessionUpdatedEventStatus
 import com.hedvig.memberservice.services.redispublisher.RedisEventPublisher
-import com.hedvig.memberservice.web.dto.GenericBankIdAuthenticationRequest
 import com.hedvig.resolver.LocaleResolver
 import org.axonframework.commandhandling.gateway.CommandGateway
 import org.springframework.beans.factory.annotation.Value
-import org.springframework.core.env.Environment
 import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Service
 import java.util.*
@@ -44,12 +42,11 @@ class ZignSecBankIdService(
 
     fun authenticate(
         memberId: Long,
-        request: GenericBankIdAuthenticationRequest,
-        zignSecAuthenticationMarket: ZignSecAuthenticationMarket,
+        personalNumber: String?,
+        market: ZignSecAuthenticationMarket,
         acceptLanguage: String?
     ): StartZignSecAuthenticationResult {
-        if (zignSecAuthenticationMarket == ZignSecAuthenticationMarket.NORWAY &&
-            request.personalNumber == null) {
+        if (market == ZignSecAuthenticationMarket.NORWAY && personalNumber == null) {
             return StartZignSecAuthenticationResult.StaticRedirect(
                 resolveNorwegianLoginUrl(memberId, acceptLanguage)
             )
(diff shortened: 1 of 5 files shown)


member-service/src/main/java/com/hedvig/memberservice/web/AuthController.kt
```diff
@@ -60,12 +62,35 @@ class AuthController @Autowired constructor(
...
+            },
+            acceptLanguage = acceptLanguage,
+            authorization = authorization.replace("Bearer ", "")
+        )
+
+        return when (result) {
+            is StartZignSecAuthenticationResult.Success ->
+                ResponseEntity.ok(ZignSecStartDto(result.redirectUrl))
+            is StartZignSecAuthenticationResult.StaticRedirect ->
+                ResponseEntity.ok(ZignSecStartDto(result.redirectUrl))
```
meletios, line 89:
```suggestion
            is StartZignSecAuthenticationResult.Success, StartZignSecAuthenticationResult.StaticRedirect ->
                ResponseEntity.ok(ZignSecStartDto(result.redirectUrl))
```

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
```diff
@@ -117,9 +116,13 @@ class ZignSecBankIdService(
...
-    private fun resolveNorwegianLoginUrl(memberId: Long, acceptLanguage: String?) = when (resolveLocaleFromMember(memberId, acceptLanguage)?.language) {
-        Locale("nb").language -> baseUrl + NORWEGIAN_BANK_ID_NORWEGIAN_LOGIN_PATH
-        else -> baseUrl + NORWEGIAN_BANK_ID_ENGLISH_LOGIN_PATH
+    private fun resolveNorwegianLoginUrl(
+        memberId: Long,
+        acceptLanguage: String?,
+        authorization: String
+    ) = when (resolveLocaleFromMember(memberId, acceptLanguage)?.language) {
+        Locale("nb").language -> "$baseUrl$NORWEGIAN_BANK_ID_NORWEGIAN_LOGIN_PATH?auth_token=$authorization"
+        else -> "$baseUrl$NORWEGIAN_BANK_ID_ENGLISH_LOGIN_PATH?auth_token=$authorization"
```
meletios, line 125: @sampettersson @styrelius is the auth token expected like this?

meletios (commented):
Nice 🙌  I suggest an improvement + invite sam and Ellen to the party

member-service/src/main/java/com/hedvig/memberservice/web/AuthController.kt
```diff
@@ -60,12 +62,35 @@ class AuthController @Autowired constructor(
...
+
+        val result = zignSecBankIdService.authenticate(
+            memberId = memberId,
+            personalNumber = request.personalNumber,
+            market = when (country) {
+                BankIdAuthCountry.norway -> ZignSecAuthenticationMarket.NORWAY
+                BankIdAuthCountry.denmark -> ZignSecAuthenticationMarket.DENMARK
+            },
+            acceptLanguage = acceptLanguage,
+            authorization = authorization.replace("Bearer ", "")
```
bystam, line 82: This is the thing we'll need to verify works as expected

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
meletios: btw i remember that it should be like `#auth_token=$authorization` but they will know

bystam: Oh - I had no idea there was already an expected formatting. I just tossed it in and thought the other side was yet to be built 😅 

Thanks for pointing it out

member-service/src/main/java/com/hedvig/memberservice/web/AuthController.kt
bystam: This does not compile. Kotlin is not cool enough to create an intersection of the APIs of `Success` and `StaticRedirect` based on that match. Then it only receives the API of its parent: `StartZignSecAuthenticationResult`, which does not have `redirectUrl`.

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
sampettersson: It's already implemented web-side so it should be attached like `#token=$authorization` :)

sampettersson: And a hash-bang is so that it's never sent through the HTTPS request on the app-side when it starts a WebView, a hashbang is neatly only accessible client javascript side 😄

member-service/src/main/java/com/hedvig/memberservice/web/AuthController.kt
sampettersson: @bystam another thing, you probably need to URLEncode the string aswell as it contains equals signs and stuff 😅

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
bystam: PR updated :)

member-service/src/main/java/com/hedvig/memberservice/web/AuthController.kt
bystam: Ah right - will do

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
styrelius: Sorry for being late to the party but seems to me like you guys have everything under control here 👌

member-service/src/main/java/com/hedvig/memberservice/web/AuthController.kt
bystam: Done!

sampettersson (approved):
to the moon 🚀

PR #469, opened 2021-02-04
read simple sign auth token from body instead of header

bystam:
# Jira Issue: []

## What?
Authorization header is implicitly stripped in api gateway. Read from body instead.

## Why?
Current code doesn't work 🙃 

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [x] Codescouted

### DIFF
--- a/member-service/src/main/java/com/hedvig/memberservice/web/AuthController.kt
+++ b/member-service/src/main/java/com/hedvig/memberservice/web/AuthController.kt
@@ -65,7 +65,6 @@ class AuthController @Autowired constructor(
     private fun auth(
         @RequestHeader("hedvig.token") memberId: Long,
         @RequestHeader(value = "Accept-Language", required = false) acceptLanguage: String?,
-        @RequestHeader("Authorization") authorization: String,
         @PathVariable("country") country: BankIdAuthCountry,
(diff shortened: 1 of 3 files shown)


member-service/src/main/java/com/hedvig/memberservice/web/AuthController.kt
```diff
@@ -79,7 +78,7 @@ class AuthController @Autowired constructor(
                 BankIdAuthCountry.denmark -> ZignSecAuthenticationMarket.DENMARK
             },
             acceptLanguage = acceptLanguage,
-            authorization = authorization.replace("Bearer ", "")
+            token = request.token
```
meletios, line 81: Do we need to encode it? filter out something as the `Bearer`?

meletios (approved):
Sweet

member-service/src/main/java/com/hedvig/memberservice/web/AuthController.kt
bystam: It's encoded inside the service (where the URL is built).

Bearer is not needed, it is attached prior to adding `Bearer` in giraffe. See https://github.com/HedvigInsurance/giraffe/pull/1055

PR #475, opened 2021-03-03
Bug fix for `Combo is not created correctly`

FredrikAreschoug:
# Jira Issue: [HVG-1003]

## What?
This was kind of a race condition so it worked 100% of the time when using bank id since that always took longer time for the member to finish that process than for `MemberSignedSaga` to pick the event up. But with `SimpleSign` it was more of a question of what happens first storing to `UnderwriterSignSessionRepository` or picking it up by `MemberSignedSaga`. 

So I had to switch how the underwriter sign session is created

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted

[HVG-1003]: https://hedvig.atlassian.net/browse/HVG-1003


meletios (approved):
Nice ✌️

PR #477, opened 2021-03-16
improve the logback config to hopefully work better in datadog

bystam:
# Jira Issue: [HVG-1050]

## What?
-

## Why?
-

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted

[HVG-1050]: https://hedvig.atlassian.net/browse/HVG-1050

### DIFF
--- a/member-service/src/main/resources/logback-spring.xml
+++ b/member-service/src/main/resources/logback-spring.xml
@@ -36,6 +36,7 @@
       <providers>
         <arguments/>
         <mdc/>
+        <logstashMarkers/>
         <stackTrace/>
         <timestamp>
           <timeZone>UTC</timeZone>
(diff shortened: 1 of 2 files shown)


member-service/src/main/resources/logback-spring.xml
```diff
@@ -85,24 +54,22 @@
...
+  <springProfile name="development">
     <root level="INFO">
       <appender-ref ref="CONSOLE"/>
-      <!--<appender-ref ref="FILE"/>-->
-      <!-- <appender-ref ref="JSON_FILE"/> -->
     </root>
   </springProfile>
 
-  <springProfile name="kubernetes">
+  <springProfile name="!development">
```
vonElfvin, line 64: Doesn't this mean that we'll have this one if we don't have a profile at all?

bystam: Well - yes. But I don't see how that's a problem? If development isn't set, we assume JSON logging (which I think is a good default). Console logging makes sense only for development purposes.

FredrikAreschoug (approved):
:runner:

FredrikAreschoug (approved):

PR #478, opened 2021-03-22
remove some MDC entries that were added in debugging panic, and which are never cleared

bystam:
# Jira Issue: []

## What?
Remove some `MDC.put(...)` entries.

## Why?
They were never cleared, which makes them bleed out to other log statements happening on the same thread. This is bad, as it can falsely tag logs with the wrong member id.

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


mightyost (approved):
Oki

PR #480, opened 2021-03-31
ME-49: Co-insured login / signup [Part 3]

bystam:
# Jira Issue: [BFF-49]

## What?
- [ME-67]: Change authentication endpoints to use new credentials tables to authenticate
- [ME-69]: Check SimpleSignConnection when no user is found

[BFF-49]: https://hedvig.atlassian.net/browse/BFF-49
[BFF-67]: https://hedvig.atlassian.net/browse/BFF-67
[BFF-69]: https://hedvig.atlassian.net/browse/BFF-69

[ME-67]: https://hedvig.atlassian.net/browse/ME-67
[ME-69]: https://hedvig.atlassian.net/browse/ME-69


member-service/src/main/java/com/hedvig/auth/services/UserService.kt
```diff
@@ -0,0 +1,110 @@
...
+     * credentials are found a new user is created with the credentials provided. The new user
+     * becomes associated with the member that you specify in the `onboardingMemberId`
+     * parameter. The next time you call `findOrCreateUserWithCredentials()` with the same
+     * credentials, the previously created user is returned.
+     *
+     * @param credentials The credentials used to lookup the user.
+     * @param onboardingMemberId If a new user needs to be created it will be associated with this member.
+     * @return A User matching the credentials, or null if the credentials are invalid.
+     */
+    fun findOrCreateUserWithCredentials(credentials: Credentials, onboardingMemberId: MemberId): User? {
```
bystam, line 45: I think the `onboarding` prefix on this parameter creates confusion where it doesn't need to be. It's true that the "member" constructs play a large role in onboarding, but he member IDs are also somewhat ubiquitous identifiers for "users/humans" throughout the entire system.

Explicitly attaching "onboarding" here makes it sound like the ID you send here has more specific constraints on itself than any other `memberId`, but I don't think that's the case. Arguably any valid `memberId` can be sent in here to run this method, so I think the prefix is misleading.

member-service/src/main/java/com/hedvig/auth/services/UserService.kt
```diff
@@ -0,0 +1,110 @@
...
+                )
+                user
+            }
+        }
+    }
+
+    private fun findOrCreateZignSecUser(credentials: Credentials.ZignSec, onboardingMemberId: MemberId): User? {
+        if (credentials.idProviderPersonId == null) {
+            return null
+        }
```
bystam, line 81: If `idProviderPersonId` is a requirement to actually get something here I don't think it should be nullable in the credentials.

member-service/src/main/java/com/hedvig/auth/services/UserService.kt
```diff
@@ -0,0 +1,110 @@
...
+            zignSecCredential.user
+        } else {
+            if (credentials.personalNumber != null) {
+                val simpleSignConnection = simpleSignConnectionRepository.findByPersonalNumberAndCountry(
+                    personalNumber = credentials.personalNumber,
+                    country = credentials.countryCode
+                )
+                if (simpleSignConnection != null) {
+                    return simpleSignConnection.user
+                }
```
bystam, line 96: In this case I think we should be creating a `ZignSecCredential` for this user so that we have authenticated credentials beyond the first login.

member-service/src/main/java/com/hedvig/auth/services/UserService.kt
```diff
@@ -0,0 +1,110 @@
...
+        )
+        return if (swedishBankIdCredential != null) {
+            swedishBankIdCredential.user
+        } else {
+            val simpleSignConnection = simpleSignConnectionRepository.findByPersonalNumberAndCountry(
+                personalNumber = credentials.personalNumber,
+                country = "SE"
+            )
+            if (simpleSignConnection != null) {
+                simpleSignConnection.user
```
bystam, line 64: SimpleSign is not currently part of the Swedish product, so there's no business requirement that says we need to support it. I guess we could for solution integrity purposes, but it could also be worthwhile avoiding the complexity/security implications?

member-service/src/main/java/com/hedvig/auth/services/UserService.kt
```diff
@@ -0,0 +1,110 @@
...
+            } else {
+                val user = User(associatedMemberId = onboardingMemberId)
+                swedishBankIdCredentialRepository.saveAndFlush(
+                    SwedishBankIdCredential(
+                        personalNumber = credentials.personalNumber,
+                        user = user
+                    )
+                )
+                user
```
bystam, line 73: As I mentioned before, when we actually create a new user for a recently generated `memberId`, I think the rest of the system is going to require us to populate the member object one way or another. Maybe we should proceed with doing that on a lazy as-needed basis, to make sure we add the minimal amount of info.

Regardless, my guess is that there are bits of the system that will fail as they will assume that a "member" that has successfully logged in will contain some minimal amount of information. And my guess is that said information is either all or some combination of:
- Names
- Birth date
- SSN/personal number

bystam: What I mean is, I think what we call member lives multiple lives. The "onboarding" side of member has everything as nullable and does not require any state.

But the "logged in user" side of member has an implicit contract on its minimal state.

bystam (commented):
Nice additions! I added a list of comments and thoughts though. Some of them are just low level code related, but I think some of them are actual things we might need to address.

member-service/src/main/java/com/hedvig/auth/services/UserService.kt
niho: Well... I figured that the `MemberSimpleSignedEvent` is defined like this:

```kotlin
data class MemberSimpleSignedEvent(
    val memberId: Long,
    val nationalIdentification: String,
    val nationality: Nationality,
    val referenceId: UUID
) {
    enum class Nationality {
        SWEDEN,
        NORWAY,
        DENMARK
    }
}
```

So, since we have the possibility that `nationality` can be `SWEDEN`, then we obviously need to handle that. And it's just strange to not be consistent.

member-service/src/main/java/com/hedvig/auth/models/SimpleSignConnection.kt
```diff
@@ -15,7 +15,7 @@ import javax.persistence.UniqueConstraint
 @Table(
     uniqueConstraints = [UniqueConstraint(columnNames = ["personal_number", "country"])]
 )
-class SimpleSignConnection(
+internal class SimpleSignConnection(
```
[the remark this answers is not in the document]
niho: I want to be explicit about what is a public API and what is not. It just forces better API design and indicates intent.

member-service/src/main/java/com/hedvig/auth/models/User.kt
```diff
@@ -9,22 +9,24 @@ import javax.persistence.Id
 import javax.persistence.OneToOne
 import javax.persistence.Table
 
+typealias MemberId = String
```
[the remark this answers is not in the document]
niho: This is also about API design and indicating intent and code readability. I'm saying that what I care about is a "member identifier", I don't care about the implementation details (it being a string). I think this helps to clarify the confusion caused by a member ID being a Long outside of our module and a String inside. It's all about someone reading the code in 5 years and trying to understand if something is random or has been done with consideration and intent.

member-service/src/test/java/com/hedvig/auth/import/SimpleSignMemberImporterTest.kt
```diff
@@ -17,10 +17,14 @@ internal class SimpleSignMemberImporterTest @Autowired constructor(
...
-        userRepository,
-        simpleSignConnectionRepository
-    )
+    private lateinit var importer: SimpleSignMemberImporter
+
+    @BeforeEach
+    fun setup() {
+        importer = SimpleSignMemberImporter()
+        importer.userRepository = userRepository
+        importer.simpleSignConnectionRepository = simpleSignConnectionRepository
```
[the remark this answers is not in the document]
niho: Yes, making them private and using `@Autowired` would be great. But it doesn't work with `@DataJpaTest`, since it will only autowire classes annotated with `@Repository`.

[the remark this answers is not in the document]
niho: ...and no-one seems to have seen the usefulness in testing services and added a `@ServiceTest` annotation to Spring. 😕

member-service/src/main/java/com/hedvig/auth/services/UserService.kt
niho: This is once again about API design and intent! From a purely technical point of view you are of course correct that it is possible to provide any good ol' member id here. But that is not the intent of the API. The API is specifically about creating a user from an onboarding state on first authentication (i.e. what you would normally call a "signup"). So prefixing this with "onboarding" just makes that a bit more explicit.

member-service/src/main/java/com/hedvig/auth/services/UserService.kt
niho: Yeah, but unfortunately it is nullable in the ZignSec API, so this check needs to happen somewhere. Basically this is the case when someone ZignSec to login using Swedish BankID (which of course is not possible today in practice, but still possible from an API point-of-view).

member-service/src/main/java/com/hedvig/auth/services/UserService.kt
niho: Yeah, that might be the case. But I feel that is separate from what we are doing in this PR. Your guess *might* be true, but we don't know for sure until we dive into that part of the problem. And we can also potentially solve it by simply changing the code that makes problematic assumptions about what a "member" is.

member-service/src/main/java/com/hedvig/auth/services/UserService.kt
niho: Yes, that's a good point. I didn't consider that...

But when thinking about it, it feels like we shouldn't "escalate" a SimpleSign user to something that it is not. The SimpleSign fallback is basically saying (from the point of view of the system): We didn't find any user for you, but you might be associated with this random unauthenticated contract that we have on file, so we are going to show you that until we know better what's going on.

niho: But I might be off on this, so please prove me wrong! :)

member-service/src/main/java/com/hedvig/auth/import/ZignSecMemberImporter.kt
```diff
@@ -14,16 +15,17 @@ import org.axonframework.config.ProcessingGroup
...
-    private val userRepository: UserRepository,
-    private val zignSecCredentialRepository: ZignSecCredentialRepository,
-    private val objectMapper: ObjectMapper
-) {
+class ZignSecMemberImporter {
 
+    @Autowired internal lateinit var userRepository: UserRepository
+    @Autowired internal lateinit var zignSecCredentialRepository: ZignSecCredentialRepository
+
+    private val objectMapper: ObjectMapper = ObjectMapper().registerKotlinModule()
```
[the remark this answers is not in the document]
niho: Is `ObjectMapper` some kind of singleton thing? I just assumed it was a simple decoder class that could be instantiated without causing any trouble.

member-service/src/main/java/com/hedvig/auth/services/UserService.kt
bystam: Right - in that case I think that check should happen outside of it. If the method is a no-op when sending null I think it's strange for it to accept null.

member-service/src/main/java/com/hedvig/auth/models/User.kt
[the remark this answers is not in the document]
bystam: Right - I get that. I just don't think it this is 100% true:
>  I don't care about the implementation details (it being a string)
since its serialisation format is relevant. As in, `User` actually needs it to be a String here, and that it never changes.

But of course, the place where it is serialised is actually in the same file where it's declared. So maybe it makes sense.

member-service/src/main/java/com/hedvig/auth/import/ZignSecMemberImporter.kt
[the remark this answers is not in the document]
bystam: Not singleton per se, but it comes with a lot of modules you can attach to it. And in other places of the application we parse the same data type (`ZignSecNotificationRequest`) with a different mapper, I think it maybe it might be confusing to mix and match.

For instance, the way you have it configured right now won't work if the data type started containing `java.time` types, because that module isn't added to it. Using the one in the spring application context makes sure we have the same modules attached to it as the rest of the application.

The one we run in production probably has about 6 different modules, but the one declared in this file now only has 1.

member-service/src/main/java/com/hedvig/auth/services/UserService.kt
bystam: > And we can also potentially solve it by simply changing the code that makes problematic assumptions about what a "member" is.

That could be changing most of the Hedvig codebase though. And then we will get into a discussion about what the "primary role" of the member obejct is: is it the onboarding state or is it the human identity beyond that. If we change all the code in the codebase to always assume a member could be an empty husk, then we will have another kind of problem.

> But I feel that is separate from what we are doing in this PR
Well - if we deploy it like this and create member<=>user connections with no data on the member, don't we risk creating corrupted data we're forced to repair afterwards?

member-service/src/main/java/com/hedvig/auth/services/UserService.kt
bystam: I am not entirely sure either. I kind of thought that simple sign-members could be considered something that's in some kind of "limbo state", where they're later promoted to the real world by proving their existance using ZignSec. And then beyond that point they would no longer be allowed to be logged in to using simple sign credentials.

But I'm not sure there's a good reason for that, other than my hunch. 🤷‍♀️

member-service/src/test/java/com/hedvig/auth/import/SimpleSignMemberImporterTest.kt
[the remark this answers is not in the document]
bystam: Right okay - but then what was the problem with the previous solution where they were constructor injected?

member-service/src/main/java/com/hedvig/auth/models/SimpleSignConnection.kt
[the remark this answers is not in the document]
bystam: Right - that makes sense. I think my brain sort of filters it out as redundant text, but that's probably because I am used to writing Swift, where internal is default.

member-service/src/main/java/com/hedvig/auth/import/ZignSecMemberImporter.kt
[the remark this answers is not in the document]
niho: Ok. So its more like some kind of global configuration? Then it might even be a good thing to use a specific instance here, since we want this stuff to be as modular as possible.

member-service/src/main/java/com/hedvig/auth/services/UserService.kt
bystam: > And then we will get into a discussion about what the "primary role" of the member obejct is

And by "we" I don't mean you and I. I think that's a debate that would have to be held broadly in the company to spread the knowledge of changes to that currently implicit contract.

member-service/src/test/java/com/hedvig/auth/import/SimpleSignMemberImporterTest.kt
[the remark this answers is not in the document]
niho: The problem (in my opinion) was that you then need to make them public and by doing that you are exposing an internal implementation detail.

I'm just trying to wage a silent war against the "dependency injection" and "mock everything" culture that I think is really bad design/practice in general.

member-service/src/main/java/com/hedvig/config/AxonConfig.java
```diff
@@ -96,6 +96,12 @@ public void configure(EventProcessingConfiguration config) {
           x ->
               TrackingEventProcessorConfiguration.forSingleThreadedProcessing()
                   .andInitialTrackingToken(StreamableMessageSource::createTailToken));
+
+      config.registerTrackingEventProcessor(
+          "MemberToUserImporter",
+          x ->
+              TrackingEventProcessorConfiguration.forSingleThreadedProcessing()
+                  .andInitialTrackingToken(StreamableMessageSource::createTailToken));
```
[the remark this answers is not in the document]
niho: It must have sneaked back somehow. I distinctly remember removing it.

member-service/src/test/java/com/hedvig/auth/import/SimpleSignMemberImporterTest.kt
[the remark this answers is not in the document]
bystam: I agree that "mock everything" is a bad practice. I think I just mostly prefer the aesthetics of placing dependencies in the constructors as opposed to having to write the `@Autowired internal lateinit var` prefix.

But yes - in a nice future where our tests simply boot the real app with an in-memory database and fire requests to test the contract I agree that this would be the nicest approach! So let's keep it like this.

member-service/src/main/java/com/hedvig/auth/import/ZignSecMemberImporter.kt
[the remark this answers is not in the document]
bystam: Right - in that case I **think** it might be more resilient to call `.findAndRegisterModules()` to make it auto-discover modules on the classpath. Otherwise it might start failing if someone later adds an `Instant` or `LocalDate` to the type we're deserialising with it, for instance.

member-service/src/main/java/com/hedvig/config/AxonConfig.java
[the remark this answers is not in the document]
bystam: Probably happened on a different branch (the one where you split things up), and then the rebase auto-merged them into this again as it looks like distinct changes

zign-sec/src/main/java/com/hedvig/external/zignSec/repository/entitys/Identity.kt
```diff
@@ -38,4 +37,26 @@ class Identity(
...
+                    this.idProviderName == other.idProviderName &&
+                    this.identificationDate == other.identificationDate &&
+                    this.idProviderRequestId == other.idProviderRequestId &&
+                    this.idProviderPersonId == other.idProviderPersonId &&
+                    this.customerPersonId == other.customerPersonId
+            }
+            else ->
+                return false
+        }
+    }
```
bystam, line 61: This is missing a `hashCode()` implementation, which is a breach of the java `Object` [contract](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object)).

I suggest using the intellij auto-generation of these methods though. It's a simple way to make sure they become correct and provide equal cover.

But also - when is this used? I see one comparison towards `null` - but I don't think that requires implementing an equality check

bystam: I'm just gonna auto-generate this and replace it. :)

niho (approved):

PR #481, opened 2021-04-06
ME-49: Co-insured login / signup [Part 2]

bystam:
# Jira Issue: [BFF-65]

## What?
- Implements an Axon event processor based export of "signed members" to the new `User` model and its login credential types

## Why?
- Our existing model/proxy for "user that can login" is our "signed member entities" (Sweden) and our "ZignSec authentication entities" (Norway/Denmark)
- We wish to have a unified "user" model that's somewhat decoupled from members
- Since we have existing signed members matching the above entities, we need to migrate them to the new model
- We need to support both migration of historical data, and population on subsequent additions

## Not in this PR
Disregard what's written below. This will be solved with this approach: https://github.com/HedvigInsurance/member-registry-service/pull/482

> Members who have signed with SimpleSign will not be exported through this, since they don't have their personal number verified through the process of signing up.

> To make sure we only create `User`s for people who verifyably own their credentials, simple signed members will have to have their users created lazily upon first login. At that time, we will make sure to establish the member<=>user connection when we have a verified personal number.

> We could probably do some cross-referencing between the `MemberSimpleSignedEvent`, `SignedMemberEntity` and `ZignSecAuthenticationEntity` to find members who both signed with simple sign and then logged in at least once. But these won't solve all cases - since there might be members who simple-signed but never logged in. Those in the last category will need the lazy user creation anyway, so it doesn't really make sense to migrate them beforehand when we can leverage the lazy solution for all simple signed members.

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [x] Codescouted

## Screenshots
After running locally, here is some example data:
image
-----
image

[BFF-65]: https://hedvig.atlassian.net/browse/BFF-65


member-service/src/test/java/com/hedvig/auth/AuthMigrationTest.kt
```diff
@@ -0,0 +1,162 @@
...
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest
+
+@DataJpaTest
+internal class AuthMigrationTest @Autowired constructor(
+    private val userRepository: UserRepository,
+    private val signedMemberRepository: SignedMemberRepository,
+    private val zignSecAuthenticationEntityRepository: ZignSecAuthenticationEntityRepository
+) {
```
bystam, line 25: I tried modeling these tests on a bit higher level, where I wanted to map real event triggering with export behaviours. But it felt like a bit of a rabbit hole to set up the `command -> aggregate -> event -> tracking even processor` chain properly and also have it testable, since tracking event processors run asynchronously.

One simple way of doing it would've been to just start the spring application in the test, but `member-service` is missing a lot of configuration for running tests, so I couldn't get it to work without doing way too much here.

member-service/src/main/java/com/hedvig/memberservice/query/AuthModelExporter.kt
```diff
@@ -0,0 +1,88 @@
...
+    fun on(e: MemberSignedEvent) {
+        val signedMember = signedMemberRepository.findById(e.id).orElse(null)
+        if (signedMember == null) {
+            logger.info("MemberSignedEvent targeted not currently signed member ${e.id} - skipping export")
+            return
+        }
+        if (userRepository.findByAssociatedMemberId(e.id.toString()) != null) {
+            logger.info("Member ${e.id} was already migrated - skipping export")
+            return
+        }
```
bystam, line 38: It appears that same member had multiple "member signed events" - at least on staging. Several of these constraints (including this one) were added once I actually ran this locally and saw it fail.

We had some really good uniqueness constraints set up at least :)

member-service/src/main/java/com/hedvig/memberservice/query/AuthModelExporter.kt
```diff
@@ -0,0 +1,85 @@
...
+    private val signedMemberRepository: SignedMemberRepository,
+    private val userRepository: UserRepository,
+    private val objectMapper: ObjectMapper
+) {
+
+    private val logger = LoggerFactory.getLogger(javaClass)
+
+    @EventHandler
+    fun on(e: MemberSignedEvent) {
+        val signedMember = signedMemberRepository.findById(e.id).orElse(null)
```
vonElfvin, line 29:
```suggestion
        val signedMember = signedMemberRepository.findByIdOrNull(e.id)
```

member-service/src/main/java/com/hedvig/memberservice/query/AuthModelExporter.kt
```diff
@@ -0,0 +1,85 @@
...
+        if (signedMember == null) {
+            logger.info("MemberSignedEvent targeted not currently signed member ${e.id} - skipping export")
+            return
+        }
```
vonElfvin, line 33: This creates an unnecessary race condition IMO. The event happened thus the member signed I'd say

member-service/src/main/java/com/hedvig/memberservice/query/AuthModelExporter.kt
```diff
@@ -0,0 +1,85 @@
...
+            zignSecNotificationJson = e.providerJsonResponse
+        )
+    }
+
+    private fun exportZignSecAuthenticatedMember(memberId: Long, zignSecNotificationJson: String) {
+        val signedMember = signedMemberRepository.findById(memberId).orElse(null)
+        if (signedMember == null) {
+            logger.info("MemberSignedEvent targeted not currently signed member $memberId - skipping export")
+            return
+        }
```
vonElfvin, line 69: Same thing here

member-service/src/main/java/com/hedvig/memberservice/query/AuthModelExporter.kt
```diff
@@ -0,0 +1,85 @@
...
+        if (userRepository.findByAssociatedMemberId(memberId.toString()) != null) {
+            logger.info("Member $memberId was already migrated - skipping export")
+            return
+        }
+
+        val zignSecNotification = objectMapper.readValue(zignSecNotificationJson, ZignSecNotificationRequest::class.java)
+        val idProviderName = zignSecNotification.identity?.idProviderName ?: return
+        val idProviderPersonId = zignSecNotification.identity?.idProviderPersonId ?: return
```
vonElfvin, line 77: Maybe throw exception? Or at least log.error?

member-service/src/main/java/com/hedvig/memberservice/query/AuthModelExporter.kt
```diff
@@ -0,0 +1,85 @@
...
+import com.hedvig.memberservice.events.MemberSignedEvent
+import com.hedvig.memberservice.events.NorwegianMemberSignedEvent
+import org.axonframework.config.ProcessingGroup
+import org.axonframework.eventhandling.EventHandler
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Component
+
+@Component
+@ProcessingGroup("AuthModelExporter")
+class AuthModelExporter(
```
vonElfvin, line 19: Maybe?
```suggestion
class AuthEventListener(
```

member-service/src/test/java/com/hedvig/auth/AuthMigrationTest.kt
vonElfvin: I think this approach is nice 👍

vonElfvin (commented):
Clean solution, had some minor question marks :)

member-service/src/main/java/com/hedvig/memberservice/query/AuthModelExporter.kt
bystam: I kind of feel semantic meaning is lost by calling it `EventListener`? Exporting is **what** it does, event listening is **how it does it**. I guess we could combine them into `AuthModelExportingEventListener` if we really want both in there.

member-service/src/main/java/com/hedvig/memberservice/query/AuthModelExporter.kt
bystam: Is it an error? I assumed it was nullable in the model because that's a valid state?

member-service/src/main/java/com/hedvig/memberservice/query/AuthModelExporter.kt
bystam: I don't understand what the race condition is? That the signed member repository wouldn't be updated? I thought those events happened on a subscribing event processor, whereas this is a tracking one, which would mean the repositories are reliably updated before the tracking ones start?

What do you suggest? :)

bystam: The reason I check with the signed member repository is because multiple members can have been signed with the same personal number, but I only want the most recent one. So I can't trust only the event to tell me that it should become a user.

member-service/src/main/java/com/hedvig/memberservice/query/AuthModelExporter.kt
bystam: I added a warning log 👍 then we can follow up whether it happened

member-service/src/main/java/com/hedvig/memberservice/query/AuthModelExporter.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    private val logger = LoggerFactory.getLogger(javaClass)
+
+    @EventHandler
+    @Transactional
+    fun on(e: MemberSignedEvent) {
+        val personalNumber = e.ssn ?: run {
+            logger.info("Member ${e.id} was signed without personal number - skipping export")
+            return
+        }
```
bystam, line 37: I had events taken from staging which didn't contain any SSN - so I had to protect against that.

member-service/src/main/java/com/hedvig/memberservice/query/AuthModelExporter.kt
```diff
@@ -0,0 +1,99 @@
...
+        swedishBankIdCredentialRepository.findByPersonalNumber(e.ssn)?.let { credential ->
+            userRepository.delete(credential.user)
+        }
```
bystam, line 40: This is an important addition. If we receive a "member signed" event for an already existing personal number, we delete the existing user for that SSN - because according to our domain, unsigning a member basically means stripping it of login rights.

This is only happening in staging, but still needs to be taken into account.

niho (commented):
Does this not include the change that moves this to the `com.hedvig.auth` package and renames it as "importer"?

bystam:
> Does this not include the change that moves this to the `com.hedvig.auth` package and renames it as "importer"?

Nope - that's done in the simple sign follow-up PR https://github.com/HedvigInsurance/member-registry-service/pull/482/files

niho (approved):
I decided to split the importer up into three parts (one for each type of member: Swedish, ZignSec and SimpleSign). Please let me know if it doesn't make any sense. Also cleaned up some stray uses of "export".

bystam:
> I decided to split the importer up into three parts (one for each type of member: Swedish, ZignSec and SimpleSign). Please let me know if it doesn't make any sense. Also cleaned up some stray uses of "export".

No I think the split up makes sense. I find it a bit strange to actively rewrite a PR someone else has written though rather than commenting on it 🤔 are you used to working this way?

niho:
> > I decided to split the importer up into three parts (one for each type of member: Swedish, ZignSec and SimpleSign). Please let me know if it doesn't make any sense. Also cleaned up some stray uses of "export".
> 
> No I think the split up makes sense. I find it a bit strange to actively rewrite a PR someone else has written though rather than commenting on it 🤔 are you used to working this way?

Sorry, maybe it's weird to do that? I'm used to working like that.

bystam:
> > > I decided to split the importer up into three parts (one for each type of member: Swedish, ZignSec and SimpleSign). Please let me know if it doesn't make any sense. Also cleaned up some stray uses of "export".
> > 
> > 
> > No I think the split up makes sense. I find it a bit strange to actively rewrite a PR someone else has written though rather than commenting on it 🤔 are you used to working this way?
> 
> Sorry, maybe it's weird to do that? I'm used to working like that.

I think doing that instead of requesting changes can probably be efficient, but I would've liked a question either here on github or on slack regarding if it's okay that you try out some ideas. Doing it without communicating first I think can a bit give the appearance that you're not happy with the PR and "decide to take matters into your own hands" - if that makes sense?

So - not opposed to collaborating on PRs like that, but would've liked the heads up. :)

PR #485, opened 2021-04-12
ME-65: Add a "simple_sign_connection" table for simple sign members

niho:
I've created a separate PR for this step, as it is independent of the importer in #481.


bystam (approved):

niho:
I'm just thinking if it's confusing to have it named "Connection"? It's not really a credential, but it's not really a connection either.

bystam:
> I'm just thinking if it's confusing to have it named "Connection"? It's not really a credential, but it's not really a connection either.

Do you have a suggestion for a better name before we merge? `SimpleSignOrigin`?

member-service/src/test/java/com/hedvig/auth/AuthTest.kt
```diff
@@ -441,4 +446,181 @@ internal class AuthTest {
...
+    fun `can delete SimpleSignConnection`() {
+        val connection = simpleSignConnectionRepository.saveAndFlush(
+            SimpleSignConnection(
+                user = User(associatedMemberId = "test-member-id"),
+                personalNumber = "1234",
+                country = "NO"
+            )
+        )
+        simpleSignConnectionRepository.delete(connection)
+        val connectionAfterDeletion = zignSecCredentialRepository.findByIdOrNull(connection.id)
```
niho, line 593: I was splitting up the `AuthTest` file in #480 and rebased that branch on this branch. I then noticed this little nasty error. The `SimpleSignConnection` test is using the `zignSecCredentialRepository`!

bystam: Oops! Will you fix it, or should I? Also did you have an opinion on the other name? SimpleSignOrigin that is.

PR #486, opened 2021-04-13
ME-49: write actual event timestamp to model to get accurate created_at date

bystam:
# Jira Issue: [ME-49]

## What?
- Make `created_at` timestamps of the auth model injectable rather than auto-generated timestamps, and send in the actual timestamp value of the originating Axon events.

## Why?
- We wish to have the created_at timestamp of these models to align with when the people behind them actually entered the information.

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [x] Codescouted

[ME-49]: https://hedvig.atlassian.net/browse/ME-49

### DIFF
--- a/member-service/src/main/java/com/hedvig/auth/import/SimpleSignMemberImporter.kt
+++ b/member-service/src/main/java/com/hedvig/auth/import/SimpleSignMemberImporter.kt
@@ -5,9 +5,11 @@ import com.hedvig.auth.model.SimpleSignConnectionRepository
 import com.hedvig.auth.model.User
 import com.hedvig.auth.model.UserRepository
 import com.hedvig.memberservice.events.MemberSimpleSignedEvent
+import java.time.Instant
 import javax.transaction.Transactional
 import org.axonframework.config.ProcessingGroup
 import org.axonframework.eventhandling.EventHandler
+import org.axonframework.eventhandling.Timestamp
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
 
@@ -22,7 +24,7 @@ class SimpleSignMemberImporter(
 
     @EventHandler
     @Transactional
-    fun on(e: MemberSimpleSignedEvent) {
+    fun on(e: MemberSimpleSignedEvent, @Timestamp timestamp: Instant) {
         if (userRepository.findByAssociatedMemberId(e.memberId.toString()) != null) {
             logger.info("Member ${e.memberId} was already imported - skipping")
             return
@@ -40,11 +42,15 @@ class SimpleSignMemberImporter(
             userRepository.flush()
         }
 
-        val user = User(e.memberId.toString())
+        val user = User(
+            associatedMemberId = e.memberId.toString(),
+            createdAt = timestamp
+        )
         user.simpleSignConnection = SimpleSignConnection(
             user = user,
             personalNumber = personalNumber,
-            country = country
+            country = country,
+            createdAt = timestamp
         )
         userRepository.save(user)
 
--- a/member-service/src/main/java/com/hedvig/auth/model/SimpleSignConnection.kt
+++ b/member-service/src/main/java/com/hedvig/auth/model/SimpleSignConnection.kt
@@ -1,6 +1,5 @@
 package com.hedvig.auth.model
 
-import org.hibernate.annotations.CreationTimestamp
 import java.time.Instant
 import javax.persistence.CascadeType
 import javax.persistence.Column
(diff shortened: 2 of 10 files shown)


member-service/src/main/java/com/hedvig/auth/import/SimpleSignMemberImporter.kt
```diff
@@ -22,7 +24,7 @@ class SimpleSignMemberImporter(
 
     @EventHandler
     @Transactional
-    fun on(e: MemberSimpleSignedEvent) {
+    fun on(e: MemberSimpleSignedEvent, @Timestamp timestamp: Instant) {
```
vonElfvin, line 27: This is not the topic if this PR, but we could have all of these events (`MemberSimpleSignedEvent`, `MemberSignedEvent` and the ZignSec events) in the same EventListener. I personally prefer that at least.

member-service/src/main/java/com/hedvig/auth/model/SimpleSignConnection.kt
```diff
@@ -23,12 +22,9 @@ class SimpleSignConnection(
     @Column(name = "personal_number")
     val personalNumber: String,
     @Column(name = "country")
-    val country: String
+    val country: String,
+    val createdAt: Instant = Instant.now()
```
vonElfvin, line 26: Nice with a default :)

member-service/src/main/java/com/hedvig/auth/import/SimpleSignMemberImporter.kt
vonElfvin: Then you can have one `@ResetHandler` for flushing the whole User table (if we want that).

member-service/src/main/java/com/hedvig/auth/import/SimpleSignMemberImporter.kt
```diff
@@ -22,7 +24,7 @@ class SimpleSignMemberImporter(
...
         if (userRepository.findByAssociatedMemberId(e.memberId.toString()) != null) {
             logger.info("Member ${e.memberId} was already imported - skipping")
             return
```
vonElfvin, line 30: One of the strengths of having separate EventListeners like this is that we can reset one type of Credential at the time with separate `@ResetHandlers`. Though if we have this early return here we would not backfil them since the user would exist, but they wouldn't have the Credential. Thus the backfill would to nothing :(

vonElfvin: Or if we have everything in one Projection with a `userRepostiry.deleteAll()` in the `@ResetHandler` we'd not hit this issue.

member-service/src/main/java/com/hedvig/auth/import/SimpleSignMemberImporter.kt
vonElfvin: This way however we can have separate `@ResetHandler`s though which is cool, that would result in a bug in the backfil though given the `return` below

member-service/src/main/java/com/hedvig/memberservice/web/ResetController.kt
```diff
@@ -33,4 +34,15 @@ class ResetController(
...
+    @PutMapping("/generic/{name}")
+    fun resetGeneric(@PathVariable name: String) {
+        eventProcessingConfiguration
+            .eventProcessorByProcessingGroup(name, TrackingEventProcessor::class.java)
+            .ifPresent { trackingEventProcessor ->
+                trackingEventProcessor.shutDown()
+                trackingEventProcessor.resetTokens()
+                trackingEventProcessor.start()
+            }
+    }
```
vonElfvin, line 47: You're missing `@ResetHandler`s for the projections!

vonElfvin (approved):
Great job and clean implementation! 

The code in this PR is fine, there is an eventual bug with the Reset though that you will notice if you run it. Maybe that should be solved separately though :)

member-service/src/main/java/com/hedvig/memberservice/web/ResetController.kt
bystam: Omg this was totally checked in by mistake. I just wrote it to try and run it again. I don't see a reason to actually have this. Sorry!

member-service/src/main/java/com/hedvig/auth/import/SimpleSignMemberImporter.kt
bystam: I had it like that before, but @niho refactored it to split it up. What was your specific reason to split them up Niklas?

As for the ResetHandler. That was just a mistake, I didn't intend to actually have reset functionality for this.

member-service/src/main/java/com/hedvig/auth/import/SimpleSignMemberImporter.kt
bystam: Again - the resetting was a mistake 😅

PR #488, opened 2021-04-15
ME-49: Co-insured login

niho:
# Jira Issue: [ME-49]

## What?
- Revert "Disable creating users from `onboardingMemberId`."
- Revert "Change tests to not assume user will be created."

## Why?
- To enable users that have not signed a contract to be able to login.

[ME-49]: https://hedvig.atlassian.net/browse/ME-49


bystam (approved):
Let's do it!

PR #489, opened 2021-04-15
ME-83: Move auth system to its own module

niho:
# Jira Issue: [ME-83]

## What?
- [ME-83]: Move auth system to its own module
- ~[ME-82]: Delete user on `MemberCancellationEvent`~
- Delete user on calls to `QualityAssuranceService.unsignMember()`

## Why?
- Will make the separation between the auth system and member-service cleaner and encourage us (long term) to eventually move the code to its own micro-service and (short term) improve the modularity of member-service.

[ME-83]: https://hedvig.atlassian.net/browse/ME-83

[ME-82]: https://hedvig.atlassian.net/browse/ME-82


niho:
The build in Codefresh fails, but I suspect that is because of a cache issue that I'm unable to solve. The build works fine for me on my local machine, doing a full docker build with no cache.

I suppose the only way to get around the failed build in Codefresh is to merge with admin privileges and simply ignore the failed check. My assumption is that it will work fine building on a new branch (i.e., master when merged).

auth/pom.xml
```diff
@@ -0,0 +1,134 @@
...
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-stdlib-jdk8</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.h2database</groupId>
+            <artifactId>h2</artifactId>
+            <scope>runtime</scope>
+        </dependency>
```
bystam, line 48: I wonder if we can tag the h2-database as `<scope>test</scope>`?

auth/src/main/kotlin/com/hedvig/auth/services/UserService.kt
```diff
@@ -0,0 +1,265 @@
...
+/**
+ * The `UserService` can be used to lookup (and create or delete) a users. A user is associated
+ * with a `MemberId` and a unique set of credentials. The credentials can be used to lookup a user.
+ * To find a user with a matching credential you use the `findOrCreateUserWithCredential()`
+ * method.
+ *
+ * @see User
+ *
+ * @author Niklas Holmgren <niklas@hedvig.com>
+ * @author Fredrik Byström <fredrik.bystrom@hedvig.com>
```
bystam, line 20: "Byström" is not my name 🙃 

```suggestion
 * @author Fredrik Tõnisson-Bystam <fredrik.bystam@hedvig.com>
```

auth/src/main/resources/application.properties
```diff
@@ -0,0 +1 @@
+
```
bystam, line 1: I think this file could be deleted. It's an alternative format to `application.yml`.

member-service/src/test/resources/application.yml
```diff
@@ -8,3 +8,54 @@ spring:
     properties:
       hibernate:
         order_by.default_null_ordering: last
+
+http.client.ssl.trust-store-password: abc123
+http.client.ssl.trust-store: /Users/niklas/hedvig/dev-env/secrets/trustStore
+http.client.ssl.key-store-password: qwerty123
+http.client.ssl.key-store: /Users/niklas/hedvig/dev-env/secrets/keyStore.jks
```
bystam, line 15: `/Users/niklas/` probably isn't right. Don't we use the `DEV_ENV_DIR` environment variable here usually? Curious how/if that is set up to work at all when running on the CI.

member-service/src/test/resources/application.yml
```diff
@@ -8,3 +8,54 @@ spring:
...
+  authentication.token: 123a
+  validSigningTargetHosts: www
+redirect.authentication.successUrl: www
+redirect.authentication.failUrl: www
+hedvig.external.zignSec.eid.baseurl: www
+
+trustpilot.oauth.apikey: test
+trustpilot.oauth.secret: test
+trustpilot.oauth.username: test
+trustpilot.oauth.password: test
```
bystam, line 59: I think these trustpilot ones can be omitted. I refactored these to only be required if `trustpilot.customerio-review-links-enabled` is set to `true`.

member-service/src/test/resources/application.yml
bystam: This is probably why the branch fails to build on Codefresh.

bystam (changes requested):
Looks nice - but I had some thoughts and comments. And at least one or two things I think need be addressed.

member-service/src/main/java/com/hedvig/memberservice/users/MemberCancellationEventHandler.kt
```diff
@@ -0,0 +1,23 @@
...
+class MemberCancellationEventHandler {
+
+    @Autowired
+    private lateinit var userService: UserService
+
+    @EventHandler
+    @Transactional
+    fun on(e: MemberCancellationEvent) {
+        userService.deleteUserWithAssociatedMemberId(e.memberId.toString())
+    }
```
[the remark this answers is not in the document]
vonElfvin: I think this could be a legacy event for when we cancelled a member rather than an insurance

member-service/src/main/java/com/hedvig/config/AxonConfig.java
```diff
@@ -80,22 +80,9 @@ public void configure(EventProcessingConfiguration config) {
                   .andInitialTrackingToken(StreamableMessageSource::createTailToken));
 
       config.registerTrackingEventProcessor(
-          "SwedishMemberImporter",
+          "com.hedvig.memberservice.users",
```
vonElfvin, line 83: I think this is the default naming and singleThreadedProcessing is the default processor so if you want you can remove this, if we think we'd like to config it in the future we can of course keep it

vonElfvin (commented):
Really nice step towards decoupling authentication, especially nice that we take the time 🥳 

I added some thoughts that might be worth considering 🤷

auth/src/main/kotlin/com/hedvig/auth/services/UserService.kt
```diff
@@ -0,0 +1,265 @@
...
+    @Autowired private lateinit var swedishBankIdCredentialRepository: SwedishBankIdCredentialRepository
+    @Autowired private lateinit var zignSecCredentialRepository: ZignSecCredentialRepository
+
+    /**
+     * The following credential types are currently supported:
+     *
+     * - SwedishBankId
+     * - ZignSec
+     * - SimpleSign
+     */
```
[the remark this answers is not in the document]
vonElfvin: I think comments as these allow for automatic generation of documentation right? That would be cool to have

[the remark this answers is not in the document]
bystam: I personally sort of disagree with the "don't write comments" approach, and find it a tad... extremist. Having good names in the code is good, but there are things that I think cannot be expressed with good names.

Part of it is the fact that names are subjective - the fact that they are obvious to one person might not be true for someone else. A longer piece of documentation can usually find ways to bridge the gap by explaining more closely.

And I feel there's an overlap between the process of writing documentation and the process of writing for instance RFCs. Sometimes, when you try to document your system/API, you realise that maybe it's actually quite hard to explain. And that can lead you towards evolving your own thinking around what you're currently building.

There are also things you can explain using documentation that you simply can't with code. For instance, the intent behind a system and how you imagine it evolving. Maybe there are ways in which you wish to express that system `A` should never depend on system `B` for purposes of decoupling. It's not very very nice to express that through names like `AthatDoesntDependOnBService`- but system/service documentation does this beautifully.

> Historically we've decided to avoid comments at Hedvig since they become hard to keep up to date.

That in turn becomes a discussion whether partial documentation is better or worse than no documentation at all. This is also subjective. I find that good, but incomplete documentation is very helpful when getting to know new things. If we regularly find our documentation to evolve into being **explicitly misleading**, then I am not entirely sure that's a 100% documentation issue. Then we could arguably start asking ourselves why our things often change to make past assumptions/contracts invalid. That would arguably violate the open/closed principle.

---- 

Of course, there's also levels to it. My general approach is that if I'm building something that is quite fundamental to a service and might be the concern of several other users, then I think documentation is helpful. If I'm writing something very "high level", like an end user UI or such, I don't find it super necessary or even redundant.

In this context, that to me means that I think it's nice to document core data models (`User`) and their primary gateways to interaction (`UserService`). But if we have some very rudimentary RESTy endpoints, then maybe it doesn't help to spell out `/** Get User by ID */`.

member-service/src/main/java/com/hedvig/config/AxonConfig.java
bystam: The reason we added this was because I think I recall that the current version of Axon (not the latest one) actually makes the default event processor type a **subscribing** one, and we thought that tracking made way more sense.

auth/src/main/kotlin/com/hedvig/auth/services/UserService.kt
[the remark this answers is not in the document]
niho: It's of course a tradeoff. The approach that I have used for many years and that I think is fairly well established in the engineering community (at least in OSS) is to *always* document your **public** API and leave documentation of internal stuff to be more ad-hoc.

That approach has many benefits:
- It forces you to think clearly about boundaries in your code. What is your intended public API and what is internal implementation detail?
- It makes your public API less likely to evolve (a good thing) and thus more likely to well designed.
- At the same time it makes your internal implementation *more* likely to evolve and change (which is also a good thing).
- And as @bystam mentions, writing documentation forces you to think clearly about what you are really doing and makes your thought process clear to anyone that needs to interact with your code in the future.

member-service/src/main/java/com/hedvig/config/SSLConfig.java
```diff
@@ -17,7 +17,13 @@ private void configureSSL() {
...
-    System.setProperty("javax.net.ssl.keyStore", env.getProperty("http.client.ssl.key-store"));
-    System.setProperty("javax.net.ssl.keyStorePassword",env.getProperty("http.client.ssl.key-store-password"));
+      String keyStore = env.getProperty("http.client.ssl.key-store");
+      if (keyStore != null) {
+          System.setProperty("javax.net.ssl.keyStore", keyStore);
+      }
+      String password = env.getProperty("http.client.ssl.key-store-password");
+      if (password != null) {
+          System.setProperty("javax.net.ssl.keyStorePassword", password);
+      }
```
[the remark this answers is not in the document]
niho: I was trying to make `@SpringBootTest` work properly. This is just a remnant of that.

auth/mvnw.cmd
```diff
@@ -0,0 +1,182 @@
...
+if exist "%HOME%\mavenrc_post.bat" call "%HOME%\mavenrc_post.bat"
+if exist "%HOME%\mavenrc_post.cmd" call "%HOME%\mavenrc_post.cmd"
+:skipRcPost
+
+@REM pause the script if MAVEN_BATCH_PAUSE is set to 'on'
+if "%MAVEN_BATCH_PAUSE%" == "on" pause
+
+if "%MAVEN_TERMINATE_CMD%" == "on" exit %ERROR_CODE%
+
+exit /B %ERROR_CODE%
```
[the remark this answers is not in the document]
niho: Those got added by Intellij, so no intent from my part. But I guess it doesn't hurt to think about the modules as more or less self-contained projects?

member-service/src/test/resources/application.yml
niho: This got left in by accident when I tried to get `@SpringBootTest` to work.

auth/src/main/kotlin/com/hedvig/auth/models/SwedishBankIdCredential.kt
```diff
@@ -15,9 +16,12 @@ internal class SwedishBankIdCredential(
     @JoinColumn(unique = true)
     val user: User,
     @Column(unique = true)
-    val personalNumber: String,
-    val createdAt: Instant = Instant.now()
```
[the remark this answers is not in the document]
niho: The hack to set `createdAt` timestamp manually was only needed for the initial import. This changes it back to use a proper `@CreationTimestamp`.

auth/mvnw.cmd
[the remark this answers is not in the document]
bystam: Right - true. Specifically having a maven wrapper, though, feels like a bit like something that's here but might mislead one to believe that this submodule is somehow built using its own maven version, which I'm pretty sure it definitely won't be.

auth/src/main/kotlin/com/hedvig/auth/services/UserService.kt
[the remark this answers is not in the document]
vonElfvin: All good points, as I said I'd prefer to have discussions with the whole backend team on the topic so we can evolve on this topic as a company. Last time we discussed it was like 1.5 years ago

member-service/src/main/java/com/hedvig/config/AxonConfig.java
vonElfvin: Ah yea that's a really good catch! I completely forgot about that 🤦

niho:
I've addressed all the feedback and fixed the failing build (I was wrong, it wasn't a cache issue as I suspected).

So this should be ready to merge, unless there is any more feedback on the changes I've made since after the review.

bystam (dismissed):
Really nice! 🚀

bystam (approved):

bystam (approved):
Looks great, though I suspect you will need to merge the second "repair" commit and revert it here to make it build upon merging.

I turned off the "auto-dissmiss approvals upon new commits" features so that you can consider this PR approved even after the final tweaks.

niho:
> [...] I suspect you will need to merge the second "repair" commit and revert it here to make it build upon merging.

I already did that with [the revert commit](https://github.com/HedvigInsurance/member-registry-service/pull/489/commits/cb07e42bec42dc0f3626923f0bf1f361c2387fd4). Or are we thinking of different things? 😕

bystam:
> > [...] I suspect you will need to merge the second "repair" commit and revert it here to make it build upon merging.
> 
> I already did that with [the revert commit](https://github.com/HedvigInsurance/member-registry-service/pull/489/commits/cb07e42bec42dc0f3626923f0bf1f361c2387fd4). Or are we thinking of different things? 😕

That was the first one that fixed members who were created from Rapio, I'm thinking about this one I merged this morning: https://github.com/HedvigInsurance/member-registry-service/pull/493

niho:
> > > [...] I suspect you will need to merge the second "repair" commit and revert it here to make it build upon merging.
> > 
> > 
> > I already did that with [the revert commit](https://github.com/HedvigInsurance/member-registry-service/pull/489/commits/cb07e42bec42dc0f3626923f0bf1f361c2387fd4). Or are we thinking of different things? 😕
> 
> That was the first one that fixed members who were created from Rapio, I'm thinking about this one I merged this morning: #493
> 
> But maybe I'm missing something?

Oh, I see. I think I deleted that as well in the merge commit. They turned up as a conflict when I merged with master and the names was so similar so I just assumed it was the same thing. 😝

PR #490, opened 2021-04-16
MX-84: Populate user-connected member with basic data upon login

bystam:
# Jira Issue: [MX-84]

## What?
- When a user logs in, populate its member with some data given to us by the login provider
- Only first and last name for now
- No new Member-level events are triggered if the names are already equal to the existing ones we know
- Rewrite some JUnit4+Mockito to JUnit5+MockK

## Why?
- When members are created and linked to a user through the new non-signed-member login, we wish to populate its associated member with some basic data.
- We bet that adding first and last name will be enough for now.

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [x] Codescouted

### DIFF
--- a/member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
+++ b/member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
@@ -12,6 +12,7 @@ import com.hedvig.memberservice.commands.InactivateMemberCommand
 import com.hedvig.memberservice.commands.InitializeAppleUserCommand
 import com.hedvig.memberservice.commands.MemberSimpleSignedCommand
 import com.hedvig.memberservice.commands.MemberUpdateContactInformationCommand
+import com.hedvig.memberservice.commands.PopulateMemberThroughLoginDataCommand
 import com.hedvig.memberservice.commands.SelectNewCashbackCommand
 import com.hedvig.memberservice.commands.SetFraudulentStatusCommand
 import com.hedvig.memberservice.commands.SignMemberFromUnderwriterCommand
@@ -255,6 +256,11 @@ class MemberAggregate() {
             && member.birthDate != cmd.birthDate) {
             apply(BirthDateUpdatedEvent(id, cmd.birthDate))
         }
+    }
+
+    @CommandHandler
+    fun handle(cmd: PopulateMemberThroughLoginDataCommand) {
+        maybeApplyNameUpdatedEvent(cmd.givenName, cmd.surname)
     }
 
     @CommandHandler
--- a/member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
+++ b/member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
@@ -7,6 +7,7 @@ import com.hedvig.external.authentication.dto.ZignSecAuthenticationResult
 import com.hedvig.external.authentication.dto.ZignSecBankIdAuthenticationRequest
 import com.hedvig.integration.apigateway.ApiGatewayService
 import com.hedvig.memberservice.commands.InactivateMemberCommand
+import com.hedvig.memberservice.commands.PopulateMemberThroughLoginDataCommand
 import com.hedvig.memberservice.commands.ZignSecSuccessfulAuthenticationCommand
 import com.hedvig.memberservice.commands.models.ZignSecAuthenticationMarket
 import com.hedvig.memberservice.query.MemberRepository
@@ -98,15 +99,16 @@ class ZignSecBankIdService(
                 }
 
                 if (user != null) {
-                    if (result.memberId != user.associatedMemberId.toLong()) {
+                    val userMemberId = user.associatedMemberId.toLong()
+                    if (result.memberId != userMemberId) {
                         logger.info("ZignSec auth completion: MemberID mismatch, inactivating ${result.memberId}")
-                        commandGateway.sendAndWait<Void>(InactivateMemberCommand(result.memberId))
-                        apiGatewayService.reassignMember(result.memberId, user.associatedMemberId.toLong())
+                        commandGateway.sendAndWait<Unit>(InactivateMemberCommand(result.memberId))
+                        apiGatewayService.reassignMember(result.memberId, userMemberId)
                     }
                     logger.info("ZignSec auth completion: Sending success command")
-                    commandGateway.sendAndWait<Void>(
+                    commandGateway.sendAndWait<Unit>(
                         ZignSecSuccessfulAuthenticationCommand(
-                            user.associatedMemberId.toLong(),
+                            userMemberId,
                             result.id,
                             result.ssn,
                             ZignSecAuthenticationMarket.fromAuthenticationMethod(result.authenticationMethod),
@@ -114,10 +116,12 @@ class ZignSecBankIdService(
                             result.identity.lastName
                         )
                     )
+                    commandGateway.sendAndWait<Unit>(
+                        PopulateMemberThroughLoginDataCommand(userMemberId, result.identity.firstName, result.identity.lastName)
+                    )
                     logger.info("ZignSec auth completion: Publishing session to redis")
                     redisEventPublisher.onAuthSessionUpdated(result.memberId, AuthSessionUpdatedEventStatus.SUCCESS)
                 } else {
-                    logger.info("ZignSec auth completion: Publishing session to redis")
                     redisEventPublisher.onAuthSessionUpdated(result.memberId, AuthSessionUpdatedEventStatus.FAILED)
                 }
             }
--- a/member-service/src/main/java/com/hedvig/memberservice/services/v2/BankIdServiceV2.kt
+++ b/member-service/src/main/java/com/hedvig/memberservice/services/v2/BankIdServiceV2.kt
@@ -11,6 +11,7 @@ import com.hedvig.memberservice.jobs.BankIdAuthCollector
 import com.hedvig.memberservice.query.CollectRepository
 import com.hedvig.memberservice.query.CollectType
 import com.hedvig.integration.apigateway.ApiGatewayService
+import com.hedvig.memberservice.commands.PopulateMemberThroughLoginDataCommand
 import com.hedvig.memberservice.jobs.SwedishBankIdMetrics
 import com.hedvig.memberservice.services.redispublisher.AuthSessionUpdatedEventStatus
 import com.hedvig.memberservice.services.redispublisher.RedisEventPublisher
@@ -82,17 +83,23 @@ class BankIdServiceV2(
                 }
                 CollectStatus.complete -> {
                     swedishBankIdMetrics.completeBankIdV2Auth()
-                    val personalNumber = bankIdRes.completionData.user.personalNumber
+                    val bankidIdentity = bankIdRes.completionData.user
+                    val personalNumber = bankidIdentity.personalNumber
                     val user = userService.findOrCreateUserWithCredentials(
                         UserService.Credentials.SwedishBankID(personalNumber),
                         onboardingMemberId = memberId.toString()
                     )
                     if (user != null) {
-                        if (memberId != user.associatedMemberId.toLong()) {
+                        val userMemberId = user.associatedMemberId.toLong()
+                        if (memberId != userMemberId) {
                             commandGateway.sendAndWait<Any>(InactivateMemberCommand(memberId))
-                            apiGatewayService.reassignMember(memberId, user.associatedMemberId.toLong())
+                            apiGatewayService.reassignMember(memberId, userMemberId)
                         }
                         redisEventPublisher.onAuthSessionUpdated(memberId, AuthSessionUpdatedEventStatus.SUCCESS)
+
+                        commandGateway.sendAndWait<Unit>(
+                            PopulateMemberThroughLoginDataCommand(userMemberId, bankidIdentity.givenName, bankidIdentity.surname)
+                        )
                     } else {
                         redisEventPublisher.onAuthSessionUpdated(memberId, AuthSessionUpdatedEventStatus.FAILED)
                     }
--- a/member-service/src/test/java/com/hedvig/memberservice/MemberAggregateTests.kt
+++ b/member-service/src/test/java/com/hedvig/memberservice/MemberAggregateTests.kt
@@ -10,6 +10,7 @@ import com.hedvig.memberservice.commands.BankIdAuthenticationStatus
 import com.hedvig.memberservice.commands.InactivateMemberCommand
 import com.hedvig.memberservice.commands.MemberSimpleSignedCommand
 import com.hedvig.memberservice.commands.MemberUpdateContactInformationCommand
+import com.hedvig.memberservice.commands.PopulateMemberThroughLoginDataCommand
 import com.hedvig.memberservice.commands.SelectNewCashbackCommand
 import com.hedvig.memberservice.commands.StartSwedishOnboardingWithSSNCommand
 import com.hedvig.memberservice.commands.SwedishBankIdAuthenticationAttemptCommand
@@ -499,7 +500,36 @@ class MemberAggregateTests {
             )
     }
 
-    private inner class AggregateFactoryM<T> internal constructor(aggregateType: Class<T>?) : AbstractAggregateFactory<T>(aggregateType) {
+    @Test
+    fun `empty member - information is populated through login command`() {
+        val memberId = 1234L
+        fixture
+            .given(
+                MemberCreatedEvent(memberId, MemberStatus.INITIATED, Instant.now())
+            )
+            .`when`(
+                PopulateMemberThroughLoginDataCommand(memberId, "First", "Lastname")
+            )
+            .expectEvents(
+                NameUpdatedEvent(memberId, "First", "Lastname")
+            )
+    }
+
+    @Test
+    fun `empty member - information is not populated through login command if the data is the same`() {
+        val memberId = 1234L
+        fixture
+            .given(
+                MemberCreatedEvent(memberId, MemberStatus.INITIATED, Instant.now()),
+                NameUpdatedEvent(memberId, "First", "Lastname")
+            )
+            .`when`(
+                PopulateMemberThroughLoginDataCommand(memberId, "First", "Lastname")
+            )
+            .expectNoEvents()
+    }
+
+    private inner class AggregateFactoryM<T> constructor(aggregateType: Class<T>?) : AbstractAggregateFactory<T>(aggregateType) {
         override fun doCreateAggregate(aggregateIdentifier: String, firstEvent: DomainEventMessage<*>?): T {
             val m = MemberAggregate()
             m.cashbackService = cashbackService
--- a/member-service/src/test/java/com/hedvig/memberservice/MemberAggregateTests.kt
+++ b/member-service/src/test/java/com/hedvig/memberservice/MemberAggregateTests.kt
@@ -516,7 +516,7 @@ class MemberAggregateTests {
     }
 
     @Test
-    fun `empty member - information is not populated through login command if the data is the same`() {
+    fun `already named member - information is not updated populated through login command`() {
         val memberId = 1234L
         fixture
             .given(
@@ -527,6 +527,22 @@ class MemberAggregateTests {
                 PopulateMemberThroughLoginDataCommand(memberId, "First", "Lastname")
             )
             .expectNoEvents()
+    }
+
+    @Test
+    fun `already named member - information is updated if one name changes`() {
+        val memberId = 1234L
+        fixture
+            .given(
+                MemberCreatedEvent(memberId, MemberStatus.INITIATED, Instant.now()),
+                NameUpdatedEvent(memberId, "First", "Lastname")
+            )
+            .`when`(
+                PopulateMemberThroughLoginDataCommand(memberId, "NewFirst", null)
+            )
+            .expectEvents(
+                NameUpdatedEvent(memberId, "NewFirst", "Lastname")
+            )
     }
 
     private inner class AggregateFactoryM<T> constructor(aggregateType: Class<T>?) : AbstractAggregateFactory<T>(aggregateType) {
--- a/member-service/src/test/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdServiceTest.kt
+++ b/member-service/src/test/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdServiceTest.kt
@@ -8,54 +8,58 @@ import com.hedvig.external.authentication.dto.ZignSecAuthenticationResult
 import com.hedvig.external.zignSec.repository.entitys.Identity
 import com.hedvig.integration.apigateway.ApiGatewayService
 import com.hedvig.memberservice.commands.InactivateMemberCommand
+import com.hedvig.memberservice.commands.PopulateMemberThroughLoginDataCommand
 import com.hedvig.memberservice.commands.ZignSecSuccessfulAuthenticationCommand
 import com.hedvig.memberservice.commands.models.ZignSecAuthenticationMarket
 import com.hedvig.memberservice.query.MemberRepository
 import com.hedvig.memberservice.services.redispublisher.AuthSessionUpdatedEventStatus
 import com.hedvig.memberservice.services.redispublisher.RedisEventPublisher
+import io.mockk.MockKAnnotations
+import io.mockk.every
+import io.mockk.impl.annotations.MockK
+import io.mockk.verify
 import org.axonframework.commandhandling.gateway.CommandGateway
-import org.junit.Test
-import org.junit.Before
-import org.junit.runner.RunWith
-import org.mockito.Mock
-import org.mockito.Mockito.anyLong
-import org.mockito.Mockito.never
-import org.mockito.Mockito.verify
-import org.mockito.junit.MockitoJUnitRunner
-import org.springframework.core.env.Environment
 import java.time.LocalDateTime
 import java.util.*
-import org.mockito.Mockito.`when` as whenever
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
 
-@RunWith(MockitoJUnitRunner::class)
 class ZignSecBankIdServiceTest {
 
-    @Mock
+    @MockK
     lateinit var zignSecAuthentication: ZignSecAuthentication
 
-    @Mock
+    @MockK
     lateinit var commandGateway: CommandGateway
 
-    @Mock
+    @MockK
     lateinit var redisEventPublisher: RedisEventPublisher
 
-    @Mock
+    @MockK
     lateinit var apiGatewayService: ApiGatewayService
 
-    @Mock
+    @MockK
     lateinit var memberRepository: MemberRepository
 
-    @Mock
+    @MockK
     lateinit var userService: UserService
-
-    @Mock
-    lateinit var env: Environment
 
     lateinit var classUnderTest: ZignSecBankIdService
 
-    @Before
+    @BeforeEach
     fun before() {
-        classUnderTest = ZignSecBankIdService(zignSecAuthentication, commandGateway, redisEventPublisher, apiGatewayService, memberRepository, userService, "success", "fail", "https://www.hedvig.com")
+        MockKAnnotations.init(this, relaxed = true)
+        classUnderTest = ZignSecBankIdService(
+            zignSecAuthentication,
+            commandGateway,
+            redisEventPublisher,
+            apiGatewayService,
+            memberRepository,
+            userService,
+            "success",
+            "fail",
+            "https://www.hedvig.com"
+        )
     }
 
     @Test
@@ -83,20 +87,40 @@ class ZignSecBankIdServiceTest {
 
         val user = User(associatedMemberId = MEMBERS_ORIGIGINAL_ID.toString())
 
-        whenever(userService.findOrCreateUserWithCredentials(
-            UserService.Credentials.ZignSec(
-                countryCode = "NO",
-                idProviderName = "BankIDNO",
-                idProviderPersonId = "9578-6000-4-365161",
-                personalNumber = SSN
-            ), onboardingMemberId = MEMBER_ID.toString())).thenReturn(user)
+        every {
+            userService.findOrCreateUserWithCredentials(
+                UserService.Credentials.ZignSec(
+                    countryCode = "NO",
+                    idProviderName = "BankIDNO",
+                    idProviderPersonId = "9578-6000-4-365161",
+                    personalNumber = SSN
+                ), onboardingMemberId = MEMBER_ID.toString()
+            )
+        } returns user
 
         classUnderTest.completeAuthentication(result)
 
-        verify(commandGateway).sendAndWait<Any>(InactivateMemberCommand(MEMBER_ID))
-        verify(apiGatewayService).reassignMember(MEMBER_ID, MEMBERS_ORIGIGINAL_ID)
-        verify(commandGateway).sendAndWait<Any>(ZignSecSuccessfulAuthenticationCommand(MEMBERS_ORIGIGINAL_ID, RESULT_ID, SSN, ZignSecAuthenticationMarket.NORWAY, "Test", "Testsson"))
-        verify(redisEventPublisher).onAuthSessionUpdated(MEMBER_ID, AuthSessionUpdatedEventStatus.SUCCESS)
+        verify {
+            commandGateway.sendAndWait(InactivateMemberCommand(MEMBER_ID))
+        }
+        verify {
+            apiGatewayService.reassignMember(MEMBER_ID, MEMBERS_ORIGIGINAL_ID)
+        }
+        verify {
+            commandGateway.sendAndWait(
+                ZignSecSuccessfulAuthenticationCommand(
+                    MEMBERS_ORIGIGINAL_ID,
+                    RESULT_ID,
+                    SSN,
+                    ZignSecAuthenticationMarket.NORWAY,
+                    "Test",
+                    "Testsson"
+                )
+            )
+        }
+        verify {
+            redisEventPublisher.onAuthSessionUpdated(MEMBER_ID, AuthSessionUpdatedEventStatus.SUCCESS)
+        }
     }
 
     @Test
@@ -124,28 +148,37 @@ class ZignSecBankIdServiceTest {
 
         val user = User(associatedMemberId = MEMBER_ID.toString())
 
-        whenever(userService.findOrCreateUserWithCredentials(
-            UserService.Credentials.ZignSec(
-                countryCode = "NO",
-                idProviderName = "BankIDNO",
-                idProviderPersonId = "9578-6000-4-365161",
-                personalNumber = SSN
-            ), onboardingMemberId = MEMBER_ID.toString())).thenReturn(user)
+        every {
+            userService.findOrCreateUserWithCredentials(
+                UserService.Credentials.ZignSec(
+                    countryCode = "NO",
+                    idProviderName = "BankIDNO",
+                    idProviderPersonId = "9578-6000-4-365161",
+                    personalNumber = SSN
+                ), onboardingMemberId = MEMBER_ID.toString()
+            )
+        } returns user
 
         classUnderTest.completeAuthentication(result)
 
-        verify(commandGateway).sendAndWait<Any>(
-            ZignSecSuccessfulAuthenticationCommand(
-                MEMBER_ID,
-                RESULT_ID,
-                SSN,
-                ZignSecAuthenticationMarket.NORWAY,
-                "Test",
-                "Testsson"
+        verify {
+            commandGateway.sendAndWait<Any>(
+                ZignSecSuccessfulAuthenticationCommand(
+                    MEMBER_ID,
+                    RESULT_ID,
+                    SSN,
+                    ZignSecAuthenticationMarket.NORWAY,
+                    "Test",
+                    "Testsson"
+                )
             )
-        )
-        verify(apiGatewayService, never()).reassignMember(anyLong(), anyLong())
-        verify(redisEventPublisher).onAuthSessionUpdated(MEMBER_ID, AuthSessionUpdatedEventStatus.SUCCESS)
+        }
+        verify(inverse = true) {
+            apiGatewayService.reassignMember(any(), any())
+        }
+        verify {
+            redisEventPublisher.onAuthSessionUpdated(MEMBER_ID, AuthSessionUpdatedEventStatus.SUCCESS)
+        }
     }
 
     @Test
@@ -171,18 +204,65 @@ class ZignSecBankIdServiceTest {
             ZignSecAuthenticationMethod.NORWAY_WEB_OR_MOBILE
         )
 
-        whenever(userService.findOrCreateUserWithCredentials(
-            UserService.Credentials.ZignSec(
-                countryCode = "NO",
-                idProviderName = "BankIDNO",
-                idProviderPersonId = SSN
-            ), onboardingMemberId = MEMBER_ID.toString())).thenReturn(null)
+        every {
+            userService.findOrCreateUserWithCredentials(
+                UserService.Credentials.ZignSec(
+                    countryCode = "NO",
+                    idProviderName = "BankIDNO",
+                    idProviderPersonId = "9578-6000-4-365161",
+                    personalNumber = SSN
+                ), onboardingMemberId = MEMBER_ID.toString()
+            )
+        } returns null
 
         classUnderTest.completeAuthentication(result)
 
-        verify(redisEventPublisher).onAuthSessionUpdated(MEMBER_ID, AuthSessionUpdatedEventStatus.FAILED)
+        verify { redisEventPublisher.onAuthSessionUpdated(MEMBER_ID, AuthSessionUpdatedEventStatus.FAILED) }
     }
 
+    @Test
+    fun completeCompletedAuthentication_success_populatesMemberData() {
+        val result = ZignSecAuthenticationResult.Completed(
+            Identity(
+                countryCode = "NO",
+                firstName = "Test",
+                lastName = "Testsson",
+                fullName = "Test Testsson",
+                personalNumber = null,
+                dateOfBirth = "1900-01-01",
+                age = 121,
+                idProviderName = "BankIDNO",
+                identificationDate = LocalDateTime.now(),
+                idProviderRequestId = null,
+                idProviderPersonId = "9578-6000-4-365161",
+                customerPersonId = null
+            ),
+            RESULT_ID,
+            MEMBER_ID,
+            SSN,
+            ZignSecAuthenticationMethod.NORWAY_WEB_OR_MOBILE
+        )
+
+        val user = User(associatedMemberId = MEMBER_ID.toString())
+        every {
+            userService.findOrCreateUserWithCredentials(
+                UserService.Credentials.ZignSec(
+                    countryCode = "NO",
+                    idProviderName = "BankIDNO",
+                    idProviderPersonId = "9578-6000-4-365161",
+                    personalNumber = SSN
+                ), onboardingMemberId = MEMBER_ID.toString()
+            )
+        } returns user
+
+        classUnderTest.completeAuthentication(result)
+
+        verify {
+            commandGateway.sendAndWait(
+                PopulateMemberThroughLoginDataCommand(MEMBER_ID, "Test", "Testsson")
+            )
+        }
+    }
 
     @Test
     fun completeFailedAuthentication_sameMemberId_doesNotInactivateMemberAndDoesNotReassignsMember() {
@@ -193,7 +273,7 @@ class ZignSecBankIdServiceTest {
 
         classUnderTest.completeAuthentication(result)
 
-        verify(redisEventPublisher).onAuthSessionUpdated(MEMBER_ID, AuthSessionUpdatedEventStatus.FAILED)
+        verify { redisEventPublisher.onAuthSessionUpdated(MEMBER_ID, AuthSessionUpdatedEventStatus.FAILED) }
     }
 
     companion object {
--- a/member-service/src/main/java/com/hedvig/memberservice/commands/PopulateMemberThroughLoginDataCommand.kt
+++ b/member-service/src/main/java/com/hedvig/memberservice/commands/PopulateMemberThroughLoginDataCommand.kt
@@ -1,0 +1,9 @@
+package com.hedvig.memberservice.commands
+
+import org.axonframework.commandhandling.TargetAggregateIdentifier
+
+data class PopulateMemberThroughLoginDataCommand(
+    @TargetAggregateIdentifier var id: Long,
+    val givenName: String,
+    val surname: String
+)
--- a/member-service/src/main/java/com/hedvig/memberservice/commands/PopulateMemberThroughLoginDataCommand.kt
+++ b/member-service/src/main/java/com/hedvig/memberservice/commands/PopulateMemberThroughLoginDataCommand.kt
@@ -4,6 +4,6 @@ import org.axonframework.commandhandling.TargetAggregateIdentifier
 
 data class PopulateMemberThroughLoginDataCommand(
     @TargetAggregateIdentifier var id: Long,
-    val givenName: String,
-    val surname: String
+    val givenName: String?,
+    val surname: String?
 )
--- a/member-service/src/main/java/com/hedvig/memberservice/events/NameUpdatedEvent.kt
+++ b/member-service/src/main/java/com/hedvig/memberservice/events/NameUpdatedEvent.kt
@@ -1,6 +1,6 @@
 package com.hedvig.memberservice.events
 
-class NameUpdatedEvent(
+data class NameUpdatedEvent(
     override val memberId: Long,
     val firstName: String?,
     val lastName: String?
--- a/member-service/src/test/java/com/hedvig/memberservice/services/v2/BankIdServiceV2Test.kt
+++ b/member-service/src/test/java/com/hedvig/memberservice/services/v2/BankIdServiceV2Test.kt
@@ -2,40 +2,45 @@ package com.hedvig.memberservice.services.v2
 
 import com.hedvig.auth.services.UserService
 import com.hedvig.external.bankID.bankId.BankIdApi
-import com.hedvig.external.bankID.bankIdTypes.*
-import com.hedvig.external.bankID.bankIdTypes.Collect.*
+import com.hedvig.external.bankID.bankIdTypes.Collect.Cert
+import com.hedvig.external.bankID.bankIdTypes.Collect.Device
+import com.hedvig.external.bankID.bankIdTypes.Collect.User
+import com.hedvig.external.bankID.bankIdTypes.CollectRequest
+import com.hedvig.external.bankID.bankIdTypes.CollectResponse
+import com.hedvig.external.bankID.bankIdTypes.CollectStatus
+import com.hedvig.external.bankID.bankIdTypes.CompletionData
 import com.hedvig.integration.apigateway.ApiGatewayService
 import com.hedvig.memberservice.commands.InactivateMemberCommand
+import com.hedvig.memberservice.commands.PopulateMemberThroughLoginDataCommand
 import com.hedvig.memberservice.jobs.SwedishBankIdMetrics
 import com.hedvig.memberservice.query.CollectRepository
 import com.hedvig.memberservice.services.redispublisher.AuthSessionUpdatedEventStatus
 import com.hedvig.memberservice.services.redispublisher.RedisEventPublisher
+import io.mockk.MockKAnnotations
+import io.mockk.every
+import io.mockk.impl.annotations.MockK
+import io.mockk.verify
 import org.axonframework.commandhandling.gateway.CommandGateway
-import org.junit.Before
-import org.junit.Test
-import org.junit.runner.RunWith
-import org.mockito.Mock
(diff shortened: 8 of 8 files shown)


member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
```diff
@@ -98,26 +99,29 @@ class ZignSecBankIdService(
...
                             result.id,
                             result.ssn,
                             ZignSecAuthenticationMarket.fromAuthenticationMethod(result.authenticationMethod),
                             result.identity.firstName,
                             result.identity.lastName
                         )
                     )
+                    commandGateway.sendAndWait<Unit>(
+                        PopulateMemberFromLoginDataCommand(userMemberId, result.identity.firstName, result.identity.lastName)
+                    )
```
vonElfvin, line 121: Would be so awesome if we could get the ssn here as well ☹️ Might be necessary if we are to create claims that we have the member's ssn 🤔

member-service/src/main/java/com/hedvig/memberservice/commands/PopulateMemberFromLoginDataCommand.kt
```diff
@@ -0,0 +1,9 @@
+package com.hedvig.memberservice.commands
+
+import org.axonframework.commandhandling.TargetAggregateIdentifier
+
+data class PopulateMemberFromLoginDataCommand(
+    @TargetAggregateIdentifier var id: Long,
```
vonElfvin, line 6: 🙈 
```suggestion
    @TargetAggregateIdentifier val id: Long,
```

member-service/src/main/java/com/hedvig/memberservice/services/v2/BankIdServiceV2.kt
```diff
@@ -82,17 +83,23 @@ class BankIdServiceV2(
...
+                        if (memberId != userMemberId) {
                             commandGateway.sendAndWait<Any>(InactivateMemberCommand(memberId))
-                            apiGatewayService.reassignMember(memberId, user.associatedMemberId.toLong())
+                            apiGatewayService.reassignMember(memberId, userMemberId)
                         }
                         redisEventPublisher.onAuthSessionUpdated(memberId, AuthSessionUpdatedEventStatus.SUCCESS)
+
+                        commandGateway.sendAndWait<Unit>(
+                            PopulateMemberFromLoginDataCommand(userMemberId, bankidIdentity.givenName, bankidIdentity.surname)
+                        )
```
vonElfvin, line 102: Will we have ssn set on the member in this scenario? Or should we maybe send it down here 🤔

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
```diff
@@ -257,6 +257,11 @@ class MemberAggregate() {
         }
     }
 
+    @CommandHandler
+    fun handle(cmd: PopulateMemberFromLoginDataCommand) {
+        maybeApplyNameUpdatedEvent(cmd.givenName, cmd.surname)
+    }
+
```
vonElfvin, line 264: Nice with explicit Command for the use case

vonElfvin (commented):
Really nice! Added some thoughts that might be worth considering 🤷

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
bystam: Yeah, I thought about it. One idea was that this model makes it super easy to attach more and more data as we go, when we know more. When you log in, the same command will be run, but if we start firing more events in the aggregate then those will happen the next time the user signs in.

This way, we can easily add more code to the aggregate once we know more things we'll need. But we also don't need to add events simply because we guess what we'll need, but an surgically attach them when we know for sure.

What do you think about that approach? Adding them lazily that is.

vonElfvin: I see now that I misinterpreted the code first runthrough. I thought this was in the event listeners 🤦  I think it makes sense! I like that it doesn't store data that we don't know that we'll need today and it can be sorted out later 😃

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
```diff
@@ -98,26 +99,29 @@ class ZignSecBankIdService(
...
+                            userMemberId,
                             result.id,
                             result.ssn,
                             ZignSecAuthenticationMarket.fromAuthenticationMethod(result.authenticationMethod),
                             result.identity.firstName,
                             result.identity.lastName
                         )
                     )
+                    commandGateway.sendAndWait<Unit>(
+                        PopulateMemberFromLoginDataCommand(userMemberId, result.identity.firstName, result.identity.lastName)
```
vonElfvin, line 120: Maybe this should be named 
```suggestion
                        MaybePopulateMemberFromLoginDataCommand(userMemberId, result.identity.firstName, result.identity.lastName)
```
Since we don't always insert the data?

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
bystam: Yes! :) Really like the Axon model here, where we can keep firing commands and then incrementally produce more events later if we find ourselves needing it.

Right now, at least we saw the name missing in the claims chat - so let's start there!

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
```diff
@@ -98,26 +99,29 @@ class ZignSecBankIdService(
...
                         logger.info("ZignSec auth completion: MemberID mismatch, inactivating ${result.memberId}")
-                        commandGateway.sendAndWait<Void>(InactivateMemberCommand(result.memberId))
-                        apiGatewayService.reassignMember(result.memberId, user.associatedMemberId.toLong())
+                        commandGateway.sendAndWait<Unit>(InactivateMemberCommand(result.memberId))
+                        apiGatewayService.reassignMember(result.memberId, userMemberId)
                     }
                     logger.info("ZignSec auth completion: Sending success command")
-                    commandGateway.sendAndWait<Void>(
+                    commandGateway.sendAndWait<Unit>(
                         ZignSecSuccessfulAuthenticationCommand(
```
niho, line 110: Looking at this a bit more closely... we are already triggering a `ZignSecSuccessfulAuthenticationCommand`. That command in turn triggers a `MemberIdentifiedEvent` which contains things like first name, last name, personal number and nationality. That event in turn is used to build up a `IdentityEntity` projection.

I'm not sure why we have a `IdentityEntity` projection that is separate from the `Member` projection? But it seems to me like the general approach of a `ZignSecSuccessfulAuthenticationCommand` that is specific to the auth method used makes a lot of sense. Maybe we should just introduce a `SwedishBankIdSuccessfulAuthenticationCommand`?

We should perhaps also look into the `IdentityEntity` stuff in more detail. Maybe it has some obscure use that is beyond me. But it seems like an unnecessary and confusing duplication of data for no sensible reason. 🤨🤔

niho (changes requested):

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
niho: There is also already a `EditMemberInfoCommand, a `MemberUpdateContactInformationCommand`, and a `SwedishBankIdAuthenticationAttemptCommand`, that all seem to fill somewhat overlapping purposes to what we want to do here.

bystam: At a glance it looks like the `IdentityEntity` is used as a way to, similarly to `SignedMemberEntity`, keep track of the subset of members that have been directly identified using a proper authentication method, in this case ZignSec.

The data is ultimately used in a public API where you can get the identity for a member ID, and if one exists it returns it, otherwise 404. I think I recall hearing that this is used in for instance H.OPE where members get a checkmark saying that they have been properly identified.

The reason it only seems to be projected as a result of `ZignSecSuccessfulAuthenticationCommand` and not any similar Swedish one is because of SimpleSign only being used outside of Sweden.

Regardless - would adding a `SwedishBankIdSuccessfulAuthenticationCommand` be the correct approach? That wouldn't solve any of these problems if we imitate the existing `ZignSec` one, as that one doesn't actually write any data to the Member. They all relay to the Identity, which is something else. We could of course make `MemberAggregate` and `MemberEventListener` actually listen to the `MemberIdentifiedEvent` and actually update its projected state, but right now it doens't do that. So I guess what you're really saying is that maybe we should start using the existing `MemberIdentifiedEvent` to populate the member?

> But it seems like an unnecessary and confusing duplication of data for no sensible reason.
I think that wording is a bit unncessary. I assume it was written the way it was to solve a specific problem. Based on the git history I think all this happened after you and I started actually, as part of the simple sign launch.

vonElfvin: If I remember correctly the "identity concept" was built in connection to simple sign in Norway. So we don't make a claim payout to members that have used simple sign but never authenticated.

I faintly remember it being the first project you got included in at Hedvig @bystam? 

The thinking for it being a separate event is I think was partly to have clarity in the event log and allow for the ability to identify through different means than authentication. For example if we at some point wish to identify a member via Passport or other whatever other means. If that's unnecessary or not is up for discussion 🤷

bystam: > I faintly remember it being the first project you got included in at Hedvig @bystam?

No, I was doing the trustpilot stuff. I helped out deploying it and bug hunting when it accidentally broke. But specifically the `IdentityEntity` isn't anything I'm super familiar with. But what you explain here is basically what I guessed myself. :)

> So we don't make a claim payout to members that have used simple sign but never authenticated.

---- 

Regardless, I see two immediate paths here:
1. Rewrite the handler for `ZignSecSuccessfulAuthenticationCommand` to also apply member-mutating events outside of the existing Identity-related ones, and then create a similar `SwedishBankIdSuccessfulAuthenticationCommand` that does the same.
2. Leave the PR as is.

If we go down the route of option 1, it would only make sense if we are also able to populate the `IdentityEntity` entries with Swedish data as well. To me that feels like a good thing. Right now it's somewhat opaque that what we call an "identity" is only something that applies to ZignSec authenticated people. It would be nice if we could let that be a standalone thing that isn't implicitly tied to Norway and Denmark.

I also actually recall this currently being special cased in H.OPE somewhere. That the frondend currently actually has logic that says something like:
```javascript
const isMemberIdentified = data.member.isSwedish || data.identity != null
```
Maybe we have an opportunity to close that gap?

bystam: I just noticed we currently also apply a `MemberIdentifiedEvent` **every time** the corresponding member logs into the app. I am not super-familiar with event sourcing, but isn't it a good idea to avoid posting redundant events to keep the event log as compact as possible?

bystam: Another follow-up thought if we go down that route: Does it even make sense to publish `MemberIdentifiedEvent` in Sweden upon login? In Sweden, we always have the person identified when they sign an insurance. The "only reason" we identify a member in NO/DK upon login is because of SimpleSign.

Then again, if we don't attach `MemberIdentifiedEvent` to Sweden simply because we know it's not necessary due to SimpleSign, then we will again have created an implicit dependency on SimpleSign, instead of letting the lifecycle of the member and its identity stand on its own.

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
bystam: I actually disagree though. The API of the command is to set the data - the fact that it avoids firing events if it doesn't need to I think is an implementation detail of the aggregate.

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
bystam: If we wish to make this uniform and apply some improvements here in the direction Nikals is suggesting, then I would propose the following:
- Let `ZignSecSuccessfulAuthenticationCommand` generate events changing the data on the member (name, personal number etc), but only if it differs from existing information
- Create an `Identity` value inside `MemberAggregate` that we can use to only actually fire `MemberIdentifiedEvent`s if it differs from what we have right now
- Add a new `SwedishBankIdSuccessfulAuthenticationCommand` (and remove the `PopulateMemberFromLoginDataCommand`) that does the same as its ZignSec counterpart: both set data on the member and populate the `IdentityEntity` data type.

niho:
> We could of course make MemberAggregate and MemberEventListener actually listen to the MemberIdentifiedEvent and actually update its projected state, but right now it doens't do that. So I guess what you're really saying is that maybe we should start using the existing MemberIdentifiedEvent to populate the member?

I think my main point was that we already seem to have at least four different commands that all purport to update the personal data of a member in some way. I don't think adding yet another one is a good way forward.

I think the `MemberIdentifiedEvent` is a great name for an event! Seems super useful for what we are doing here. Just to summarise what we want to accomplish:

1. A user logs in.
2. Using a method that supplies the identity data of the person (e.g. name, age, etc.)
3. We want to update the member with that data.

Seems to me like **exactly** what the `MemberIdentifiedEvent` is (or at least signals with its name). It's just strange that it doesn't actually update the member, but rather creates **another** member entity in a completely separate table called `IdentityEntity`. Why not just set a flag on the already existing `MemberEntity` projection: e.g. set `hasSecureIdentity = true`, or whatever?

> If I remember correctly the "identity concept" was built in connection to simple sign in Norway. So we don't make a claim payout to members that have used simple sign but never authenticated. [...] The thinking for it being a separate event is I think was partly to have clarity in the event log and allow for the ability to identify through different means than authentication. For example if we at some point wish to identify a member via Passport or other whatever other means. If that's unnecessary or not is up for discussion 🤷

I don't get why that would require you to create an entierly new projection that is just the same thing as the `MemberEntity` projection? Why not just a flag on the member?

member-service/src/main/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdService.kt
niho: > We could of course make MemberAggregate and MemberEventListener actually listen to the MemberIdentifiedEvent and actually update its projected state, but right now it doens't do that. So I guess what you're really saying is that maybe we should start using the existing MemberIdentifiedEvent to populate the member?

I think my main point was that we already seem to have at least four different commands that all purport to update the personal data of a member in some way. I don't think adding yet another one is a good way forward.

I think the `MemberIdentifiedEvent` is a great name for an event! Seems super useful for what we are doing here. Just to summarise what we want to accomplish:

1. A user logs in.
2. Using a method that supplies the identity data of the person (e.g. name, age, etc.)
3. We want to update the member with that data.

Seems to me like **exactly** what the `MemberIdentifiedEvent` is (or at least signals with its name). It's just strange that it doesn't actually update the member, but rather creates **another** member entity in a completely separate table called `IdentityEntity`. Why not just set a flag on the already existing `MemberEntity` projection: e.g. set `hasSecureIdentity = true`, or whatever?

> If I remember correctly the "identity concept" was built in connection to simple sign in Norway. So we don't make a claim payout to members that have used simple sign but never authenticated. [...] The thinking for it being a separate event is I think was partly to have clarity in the event log and allow for the ability to identify through different means than authentication. For example if we at some point wish to identify a member via Passport or other whatever other means. If that's unnecessary or not is up for discussion 🤷

I don't get why that would require you to create an entierly new projection that is just the same thing as the `MemberEntity` projection? Why not just a flag on the member?

niho: So, just to summarise our findings on this from our meeting on Monday (for sake of posterity):

- We realised that the changes introduced in this PR would have completely broken the fraud detection feature that the `IdentityEntity` projection was meant to solve.
- We would not have realised this unless we had looked closely at how this PR overlap with the `IdentityEntity` and `MemberIdentifiedEvent` and questioned the sanity of the design. It initially felt like we might be going down a rabbit hole, but *in reality* it was an **extremely** relevant discussion to have, which we only realised in retrospect (I don't know if there is some kind of general learning in that, but probably 🧐).
- Based on input from Elvin we discussed whether or not we should try tro maintain backwards compatibility with the `IdentityEntity` feature (as currently used by Hope), or wether or not that feature is even useful at all. Maybe we can just kill all of that code? We discussed several possible contingent solutions to the problem and Elvin was tasked with investigating the question further and get back to the team ASAP with a conclusion so that we can move forward.

member-service/src/main/java/com/hedvig/memberservice/commands/AuthenticatedIdentificationCommand.kt
```diff
@@ -0,0 +1,17 @@
...
+    val lastName: String?,
+    val nationalIdentifier: String,
+    val countryCode: String,
+    val source: Source
+) {
+    sealed class Source {
+        object SwedishBankID: Source()
+        data class ZignSec(val idProviderName: String): Source()
+    }
+}
```
bystam, line 17: This is `ZignSecSuccessfulAuthenticationCommand` rewritten to fit several use cases.

member-service/src/main/java/com/hedvig/memberservice/commands/models/ZignSecAuthenticationMarket.kt
```diff
@@ -1,22 +1,16 @@
...
 
 import com.hedvig.external.authentication.dto.ZignSecAuthenticationMethod
-import com.hedvig.memberservice.events.MemberIdentifiedEvent
 
-enum class ZignSecAuthenticationMarket {
-    NORWAY,
-    DENMARK;
+enum class ZignSecAuthenticationMarket(val countryCode: String) {
+    NORWAY("NO"),
+    DENMARK("DK");
```
bystam, line 7: I felt like we were doing so very many layers of enum conversions where I find strings to be well established and usable.

Strong types have their use, but I think using them instead of just ISO country codes is a bit much.

member-service/src/main/java/com/hedvig/memberservice/events/MemberIdentifiedEvent.kt
```diff
@@ -12,14 +12,13 @@ data class MemberIdentifiedEvent(
...
-        DENMARK
-    }
+    enum class Nationality(val countryCode: String) {
+        SWEDEN("SE"),
+        NORWAY("NO"),
+        DENMARK("DK");
 
-    enum class IdentificationMethod {
-        NORWEGIAN_BANK_ID,
-        DANISH_BANK_ID
```
bystam, line 23: It appeared we were only using this as something we wrote on the object, not for any actual logic beyond this point.

I think it's better store this as some kind of string (or string wrapper if we really want to) that paints a better picture. In this PR I changed it to:
- `com.bankid` for swedish bankid (their domain)
- `com.zignsec:$idProviderName` for ZignSec (so for instance `com.zignsec:BankIDNO`)

member-service/src/main/java/com/hedvig/memberservice/query/MemberEventListener.java
```diff
@@ -271,4 +271,11 @@ void on(PickedLocaleUpdatedEvent e) {
     m.setPickedLocale(e.getPickedLocale());
     userRepo.save(m);
   }
+
+  void on(@NotNull MemberIdentifiedEvent event) {
+      MemberEntity m = userRepo.findById(event.getMemberId()).get();
+      m.setFirstName(event.getFirstName());
+      m.setLastName(event.getLastName());
+      m.setSsn(event.getNationalIdentification().getIdentification());
+  }
```
bystam, line 280: This is the actual thing solving the use case of the Jira ticket.

member-service/src/main/java/com/hedvig/memberservice/services/v2/BankIdServiceV2.kt
```diff
@@ -82,16 +83,30 @@ class BankIdServiceV2(
...
+                        commandGateway.sendAndWait<Any>(
+                            AuthenticatedIdentificationCommand(
+                                id = userMemberId,
+                                firstName = bankidIdentity.givenName,
+                                lastName = bankidIdentity.surname,
+                                nationalIdentifier = personalNumber,
+                                countryCode = "SE",
+                                source = AuthenticatedIdentificationCommand.Source.SwedishBankID
+                            )
+                        )
```
bystam, line 108: And here's the addition for Sweden, posting `AuthenticatedIdentificationCommand` here as well.

member-service/src/main/java/com/hedvig/memberservice/web/IdentityController.kt
```diff
@@ -1,25 +0,0 @@
...
-    private val identityService: IdentityService
-) {
-
-    @GetMapping("{memberId}/identity")
-    fun identity(
-        @PathVariable("memberId") memberId: Long
-    ): ResponseEntity<IdentityDto> =
-        identityService.identityOnMemberId(memberId)?.let {
-            ResponseEntity.ok(IdentityDto.from(it))
-        } ?: ResponseEntity.notFound().build()
```
bystam, line 23: This one returned an empty 404 if there was no entity, which it will still do now that the endpoint is gone entirely. This means this should have the same effect as if we were to not have any identity entities.

We should of course fix the frontend too as part of this, but I think it's safe to remove it as part of this PR.

member-service/src/main/java/com/hedvig/memberservice/events/MemberIdentifiedEvent.kt
```diff
@@ -3,7 +3,7 @@ package com.hedvig.memberservice.events
 data class MemberIdentifiedEvent(
     val memberId: Long,
     val nationalIdentification: NationalIdentification,
-    val identificationMethod: IdentificationMethod,
+    val identificationMethod: String,
```
bystam, line 6: This is a change in an event, but it goes from an enum (string in JSON) to just a string, so it should be backwards compatible.

bystam:
This PR has now been updated to reflect our most recent learnings:
- The `IdentityEntity` is something we can remove, and clean out both in Hedgehog and back-office (I will look into it)
- We now fire a new command de-coupled from ZignSec that produces a `MemberIdentified` event in all markets (but only if there's an actual change)
- The member is actually updated upon `MemberIdentified` (both aggregate and projection)

member-service/src/main/java/com/hedvig/memberservice/commands/AuthenticatedIdentificationCommand.kt
niho: This looks great. The only concern is about the name `AuthenticatedIdentificationCommand`. Why not `SuccessfulAuthenticationCommand`?

And I'm also thinking that another way to structure the command could be like this:

```kotlin
data class Command(
  @TargetAggregateIdentifier val id: Long,
  val authMethod: AuthMethod
) {
  sealed class AuthMethod {
    data class SwedishBankId(
      val completionData: com.hedvig.external.BankID.CompletionData // This includes firstName, lastName and personalNumber
    ): AuthMethod()

    data class ZignSec(
      val identity: com.hedvig.external.zignSec.ZignSecIdentity
    ): AuthMethod()
  }
}
```

...but since it's a command, it's just about semantics really. So not important in practice.

Doing it my proposed way would of course defer the interpretation of the data to the event handler, so might from that point of view be considered worse. But I kinda' like when the data is just passed along in an unchanged way for as long as possible. I think it's nice when you don't make a lot of unnecessary transformations and renaming of fields and stuff. It makes it easier to understand and follow the code. But that's just my personal aesthetic opinion.

niho (approved):

member-service/src/main/java/com/hedvig/memberservice/aggregates/Member.kt
```diff
@@ -11,5 +11,6 @@ data class Member(
     val acceptLanguage: String? = null,
     val pickedLocale: PickedLocale? = null,
     val livingAddress: LivingAddress? = null,
-    val birthDate: LocalDate? = null
+    val birthDate: LocalDate? = null,
+    val countryCode: String? = null
```
vonElfvin, line 15: We've had this debate forever at Hedvig, and the discussion wether a member is from a country or not is quite hot topic. The current consensus is that a member is not from a country, it's just insured in one. Thus a properly like this (depending on its usage) is not aligning with that consensus. We could of course bring up that discussion again if you like.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
```diff
@@ -305,43 +305,47 @@ class MemberAggregate() {
...
                 MemberIdentifiedEvent.NationalIdentification(
                     cmd.personalNumber,
-                    when (cmd.zignSecAuthMarket) {
-                        ZignSecAuthenticationMarket.NORWAY -> MemberIdentifiedEvent.Nationality.NORWAY
-                        ZignSecAuthenticationMarket.DENMARK -> MemberIdentifiedEvent.Nationality.DENMARK
-                    }
+                    MemberIdentifiedEvent.Nationality.fromCountryCode(cmd.zignSecAuthMarket.countryCode)
                 ),
-                cmd.zignSecAuthMarket.toMemberIdentifiedEventIdentificationMethod(),
+                "com.zignsec:$idProvider",
```
vonElfvin, line 321: Why "com"?

vonElfvin: Also, are we upcasting the events to match this change?

member-service/src/main/java/com/hedvig/memberservice/commands/AuthenticatedIdentificationCommand.kt
```diff
@@ -0,0 +1,17 @@
...
+    @TargetAggregateIdentifier val id: Long,
+    val firstName: String?,
+    val lastName: String?,
+    val nationalIdentifier: String,
+    val countryCode: String,
+    val source: Source
+) {
+    sealed class Source(val identifier: String) {
+        object SwedishBankID: Source("com.bankid")
+        data class ZignSec(val idProviderName: String): Source("com.zignsec:$idProviderName")
```
vonElfvin, line 15: Just a personal taste, if possible can we not mix object and data class?

member-service/src/main/java/com/hedvig/memberservice/commands/models/ZignSecAuthenticationMarket.kt
vonElfvin: There is an enum of ISO country codes called `CountryCode` provided by neovisionaries that we commonly use

member-service/src/main/java/com/hedvig/memberservice/events/MemberIdentifiedEvent.kt
vonElfvin: The thought here was that IdentificationMethod could be like "Passport" or other

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
```diff
@@ -305,43 +305,47 @@ class MemberAggregate() {
...
 
     @CommandHandler
-    fun handle(command: ZignSecSuccessfulAuthenticationCommand) {
+    fun handle(command: AuthenticatedIdentificationCommand) {
+        if (member.firstName == command.firstName &&
+            member.lastName == command.lastName &&
+            member.ssn == command.nationalIdentifier &&
+            member.countryCode == command.countryCode) return
         apply(
             MemberIdentifiedEvent(
```
vonElfvin, line 335: Why do we prefer using the "MemberIdentifiedEvent" or just apply "Name changed" and "SsnUpdated" events?

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
```diff
@@ -631,4 +635,14 @@ class MemberAggregate() {
...
     }
+
+    @EventSourcingHandler
+    fun on(e: MemberIdentifiedEvent) {
+        member = member.copy(
+            firstName = e.firstName,
+            lastName = e.lastName,
+            ssn = e.nationalIdentification.identification,
+            countryCode = e.nationalIdentification.nationality.countryCode
+        )
```
vonElfvin, line 646: It feels like we are using this event for a new concept now, that is probably fine but just so we know that we are changing the intent of the event

member-service/src/main/java/com/hedvig/memberservice/query/MemberEventListener.java
vonElfvin: This is a new usage of the event's intent, which as I said before is most likely fine. But we have events called "NameChanged" and "SsnUpdated" that are intended to update information like this. I don't see any limitation to us changing that, just so we are aware

member-service/src/main/java/com/hedvig/memberservice/services/v2/BankIdServiceV2.kt
```diff
@@ -82,16 +83,30 @@ class BankIdServiceV2(
...
                     if (user != null) {
-                        if (memberId != user.associatedMemberId.toLong()) {
+                        val userMemberId = user.associatedMemberId.toLong()
+                        if (memberId != userMemberId) {
                             commandGateway.sendAndWait<Any>(InactivateMemberCommand(memberId))
-                            apiGatewayService.reassignMember(memberId, user.associatedMemberId.toLong())
+                            apiGatewayService.reassignMember(memberId, userMemberId)
                         }
+
+                        commandGateway.sendAndWait<Any>(
```
vonElfvin, line 99:
```suggestion
                        commandGateway.sendAndWait<Unit>(
```

member-service/src/main/java/com/hedvig/memberservice/aggregates/Member.kt
vonElfvin: It feels like in a scenario like this one we can probably connect it to the actual authentication instead of adding it as a property on the Member

vonElfvin (commented):
Adding country concept to a member is a quite hot topic we have fought about a lot at Hedvig before. 

Also I think we are missing an upcaster here

member-service/src/main/java/com/hedvig/memberservice/aggregates/Member.kt
bystam: > We've had this debate forever at Hedvig, and the discussion wether a member is from a country or not is quite hot topic. The current consensus is that a member is not from a country, it's just insured in one. Thus a properly like this (depending on its usage) is not aligning with that consensus. We could of course bring up that discussion again if you like.

That's interesting. I haven't thought a lot about it, but my immediate thought is that a member already **is** somewhat connected to a country through its `ssn` value, which is a country-local identifier.

I'm not 100% sure I feel like adding country here makes it "local to a country", as much as its the country in which its identification comes from. I mean, lots of the values on the member is actually populated through its country local identity, including the names as well.

So, in a sense, I would question whether it is actually currently country-agnostic. A country-agnostic model would be one where a member has some kind of one-to-many relationship to a `NationalIdentity` type.

----

**Regardless**, this is only used in order to make sure we don't over-fire `MemberIdentifiedEvent`, but I can easily just drop `country` from that comparison :) it won't make a difference.

I'll do that, to make sure I don't re-light this discussion.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
bystam: I talked about the backwards compatibility of it in [this comment](https://github.com/HedvigInsurance/member-registry-service/pull/490#discussion_r617522941). In serialised form it was as string, so making it a string should make it still work.

Of course, old values won't give them the `com.bla` format, but from reading the code I felt that this was more of a bookeeping value than it was a logic one, since we don't actually read it anywhere.

Regardin the "com", that's just a... somewhat common pattern when writing somewhat unique identifier: https://en.wikipedia.org/wiki/Reverse_domain_name_notation

Swedish BankID and ZignSec own the domains `bankid.com` and `zignsec.com` respectively, so by adding it, the identifier avoids being ambiguous. And then I also don't have to invent any name for it, I can just use the existing, globally recognised "names" for these sources.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
bystam: Hmm - I thought we all agreed on using the `MemberIdentifiedEvent` here, because its name is so perfect: we just received proof that a person is who she/he claims to be, thus the person is identified. The person didn't change their name/ssn, they became identified.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
bystam: Again, I felt like this is what we agreed upon in our meeting, that the happening of "a member is identified through authentication" should drive the state of the member.

If we don't use this event for this, we might as well scrap it entirely, as it's not consumed by anything else anymore.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
vonElfvin: I'm thinking if the string format is irrelevant then sure it is backwards compatible but here it feels like you have logic baked in to the string thus an upcast would be healthy for future developers who might want to utilize that

member-service/src/main/java/com/hedvig/memberservice/commands/AuthenticatedIdentificationCommand.kt
bystam: Well, Kotlin does not allow data classes to be empty, so if I were to make it a class I would have to use a non-data one, which in turn won't give it equality checking if I were to write tests for it.

Regardless, the object will probably go away to cater to a suggestion by Niklas in another comment :)

member-service/src/main/java/com/hedvig/memberservice/commands/AuthenticatedIdentificationCommand.kt
bystam: > Why not SuccessfulAuthenticationCommand?

I think I mentally shifted towards this command primarily being about a person getting identified, and not about that it authenticated. But you're right that it does happen during authentication, so I will rename it back to where it was before.

> And I'm also thinking that another way to structure the command could be like this:

Yeah, I also went down that path a bit. One thing about it was that there was so much `when` code everywhere in the aggregate that way. Another was that it was a bit hard to figure out which types to import into the command. I also got stuck mentally when considering whether the command should actually know about outside domains.

`ZignSecIdentity` also isn't enough to do the ZignSec one, since it doesn't properly contain the personal number.

Regardless, I will rewrite it to your suggestion to get a feel for it. I could also build it in a way so that digging out the underlying data for the different `sealed class` cases is still done in the command, just to make the aggregate code less jarring. I'll look into it.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
bystam: Right, so you suggest that the upcaster would transform `NORWEGIAN_BANK_ID` to `com.zignsec:BankIDNO` so the data is uniform? That's a good idea. I'll add  it.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
niho: Yeah, I agree with Fredrik here. I don't see how we are using it for a new concept? The event is literally named `MemberIdentifiedEvent` and has always been emitted on authentication. We are just updating the `MemberEntity` projection instead of the `IdentityEntity` projection, which doesn't really change anything (because the `IdentityEntity` projection did not fill a purpose and we are going to delete it).

...and it seems to me like we have discussed this in depth already (see [my summary of our Monday meeting](https://github.com/HedvigInsurance/member-registry-service/pull/490#discussion_r617262781)).

niho (approved):

niho (changes requested):
Sorry, was a bit quick to approve. Thought you had already made all the suggested changes.

member-service/src/main/java/com/hedvig/memberservice/events/MemberIdentifiedEvent.kt
bystam: Right, that can definitely still be done with a string. The strength with a string is that it can retain dynamic information. I kind of feel like enums aren't really good to use for these purposes, and creating one based on the assumption that "maybe it could be passport" in the future doesn't seem like a strength to me necessarily.

If we do end up putting passport here, wouldn't it be nicer to spell it as: `passport:FR` instead of just `PASSPORT`?

member-service/src/main/java/com/hedvig/memberservice/commands/SuccessfulAuthenticationCommand.kt
```diff
@@ -0,0 +1,51 @@
...
+            val completionData: CompletionData
+        ): AuthMethod(
+            "com.bankid"
+        )
+
+        data class ZignSec(
+            val personalNumber: String,
+            val identity: ZignSecIdentity
+        ): AuthMethod(
+            "com.zignsec:${identity.idProviderName ?: "unknown"}"
```
bystam, line 41: `idProviderName` should really be non-nullable, but better safe than sorry?

member-service/src/main/java/com/hedvig/memberservice/commands/SuccessfulAuthenticationCommand.kt
```diff
@@ -0,0 +1,51 @@
...
+            nationalIdentifier = method.completionData.user.personalNumber,
+            countryCode = CountryCode.SE
+        )
+        is AuthMethod.ZignSec -> Identity(
+            firstName = method.identity.firstName,
+            lastName = method.identity.lastName,
+            nationalIdentifier = method.personalNumber,
+            countryCode = CountryCode.getByAlpha2Code(method.identity.countryCode)
+        )
+    }
```
bystam, line 26: Just to keep the MemberAggregate a bit lighter I decided to reduce the different method data into an identity in here. What do you think of this @niho

member-service/src/main/java/com/hedvig/memberservice/commands/SuccessfulAuthenticationCommand.kt
niho: Hehe, yeah. Crash or corrupt. The eternal question.

member-service/src/main/java/com/hedvig/memberservice/commands/SuccessfulAuthenticationCommand.kt
niho: Yes, I like it! It keeps the code in a single place.

And the `Identity` class can easily be moved out of this command as a single coherent chunk later, if we want to use the concept in more places in the future.

member-service/src/main/java/com/hedvig/memberservice/events/SSNUpdatedEvent.kt
```diff
@@ -0,0 +1,55 @@
...
+            }
+        }
+
+        private fun nationalityFromSsn(ssn: String): String = when (ssn.length) {
+            10 -> "DENMARK"
+            11 -> "NORWAY"
+            12 -> "SWEDEN"
+            else -> throw RuntimeException("Failed upcasting SSN to nationality, length = ${ssn.length}")
+        }
+    }
```
bystam, line 54: These were two upcasters that did the same thing except for having different `payload_type` names for source and destination. Normally I wouldn't collapse things like these into a single class, but since `OnboardingStartedWithSSNEvent` and `SSNUpdatedEvent` were in an actual inheritance relationship, I think it makes sense to have them be the same class but configured with different type names.

member-service/src/main/java/com/hedvig/memberservice/events/MemberIdentifiedEvent.kt
```diff
@@ -23,4 +29,40 @@ data class MemberIdentifiedEvent(
...
+                val translatedValue = when (val oldValue = document.rootElement.element("identificationMethod").text) {
+                    "NORWEGIAN_BANK_ID" -> "com.zignsec:BankIDNO"
+                    "DANISH_BANK_ID" -> "com.zignsec:NemID"
+                    else -> oldValue
+                }
+                document.rootElement.element("identificationMethod").text = translatedValue
+                document
+            }
+        }
+    }
```
bystam, line 66: I think it made sense to keep upcasters nested in the actual event. That way it becomes local, and one can easily connect events with how they have looked historically, for cohesion purposes.

member-service/src/main/java/com/hedvig/memberservice/events/MemberIdentifiedEvent.kt
```diff
@@ -23,4 +29,40 @@ data class MemberIdentifiedEvent(
...
+
+    private class IdentificationMethodUpcaster: SingleEventUpcaster() {
+
+        private val inputType = SimpleSerializedType(
+            "com.hedvig.memberservice.events.MemberIdentifiedEvent", null
+        )
+
+        private val outputType = SimpleSerializedType(
+            "com.hedvig.memberservice.events.MemberIdentifiedEvent", "1.0"
+        )
```
bystam, line 46: I noticed we used to do a `MyClass::class.java.typeName` for these, but I think that's actually not... the correct way to it. Doing it like that means that moving the even to a new package would break the upcaster. But the upcaster must be written to match what the state of the data was at the time it was introduced, and be unaffected by later changes.

So I think it makes more sense not to actually reference any of the event data types when writing an upcaster, but instead just spell it out in its serialised form.

member-service/src/main/java/com/hedvig/config/AxonConfig.java
```diff
@@ -88,9 +93,10 @@ public void configure(EventProcessingConfiguration config) {
...
     @Bean
     public EventUpcasterChain eventUpcasters() {
-        return new EventUpcasterChain(
-            new OnboardingStartedWithSSNEventUpcaster(),
-            new SSNUpdatedEventUpcaster()
-        );
+      List<EventUpcaster> upcasters = new ArrayList<>();
+      upcasters.addAll(OnboardingStartedWithSSNEvent.Companion.getUpcasters());
+      upcasters.addAll(SSNUpdatedEvent.Companion.getUpcasters());
+      upcasters.addAll(MemberIdentifiedEvent.Companion.getUpcasters());
```
bystam, line 99: Again, I think grouping the upcasters with their target event types makes it read quite nicely.

Of course, Axon doesn't **force** you to have one upcaster per event type, but since that's the pattern we've been doing I think it brings nice cohesion to group them like that.

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
```diff
@@ -305,43 +305,45 @@ class MemberAggregate() {
     }
 
     private fun applyMemberIdentifiedEventFromZignSecSignCommand(cmd: ZignSecSignCommand) {
```
vonElfvin, line 307: Would be interesting if we could make this reusable for different sign options in the future
```suggestion
    private fun maybeApplyMemberIdentified(identity: SignSuccessfullCommand.Identity) {
```

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
```diff
@@ -305,43 +305,45 @@ class MemberAggregate() {
...
             )
         )
     }
 
     @CommandHandler
-    fun handle(command: ZignSecSuccessfulAuthenticationCommand) {
+    fun handle(command: SuccessfulAuthenticationCommand) {
+        if (member.firstName == command.identity.firstName &&
+            member.lastName == command.identity.lastName &&
+            member.ssn == command.identity.nationalIdentifier) return
```
vonElfvin, line 331: Depending on how we treat the "MemberIdentifiedEvent" concept, maybe we can have another approach here. The code is a bit "axon-y" I could maybe elaborate a bit more over CoScreen what I mean if it's cryptic 😄 

```suggestion
        if (isIdentified) return
```

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
```diff
@@ -305,43 +305,45 @@ class MemberAggregate() {
     }
 
     private fun applyMemberIdentifiedEventFromZignSecSignCommand(cmd: ZignSecSignCommand) {
+        val request = objectMapper.readValue(cmd.provideJsonResponse, ZignSecNotificationRequest::class.java)
+        val idProvider = request.identity?.idProviderName ?: return
+        if (member.firstName == cmd.firstName &&
+            member.lastName == cmd.lastName &&
+            member.ssn == cmd.personalNumber) return
```
vonElfvin, line 312: Depending on how we treat the "MemberIdentifiedEvent" concept, maybe we can have another approach here. The code is a bit "axon-y" I could maybe elaborate a bit more over CoScreen what I mean if it's cryptic 😄
```suggestion
        if (isIdentified) return
```

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
```diff
@@ -631,4 +633,13 @@ class MemberAggregate() {
...
     }
+
+    @EventSourcingHandler
+    fun on(e: MemberIdentifiedEvent) {
+        member = member.copy(
+            firstName = e.firstName,
+            lastName = e.lastName,
+            ssn = e.nationalIdentification.identification
+        )
+    }
```
vonElfvin, line 644:
```suggestion
    fun on(e: MemberIdentifiedEvent) {
        member = member.copy(
            firstName = e.firstName,
            lastName = e.lastName,
            ssn = e.nationalIdentification.identification
        )
        this.isIdentified = true
    }
```

vonElfvin: Can add `isIdentified = true` to multiple events if we want

member-service/src/main/java/com/hedvig/memberservice/commands/SuccessfulAuthenticationCommand.kt
```diff
@@ -0,0 +1,51 @@
...
+import com.hedvig.external.zignSec.client.dto.ZignSecIdentity
+import com.neovisionaries.i18n.CountryCode
+import org.axonframework.commandhandling.TargetAggregateIdentifier
+
+data class SuccessfulAuthenticationCommand(
+    @TargetAggregateIdentifier val id: Long,
+    val method: AuthMethod
+) {
+
+    val identity: Identity = when (method) {
```
vonElfvin, line 13: I'm not 100% what the difference is here really but this makes sense to me
```suggestion
    val identity: Identity
        get() = when (method) {
```

member-service/src/main/java/com/hedvig/memberservice/events/MemberIdentifiedEvent.kt
```diff
@@ -12,14 +20,49 @@ data class MemberIdentifiedEvent(
...
+        override fun doUpcast(intermediateRepresentation: IntermediateEventRepresentation): IntermediateEventRepresentation {
+            return intermediateRepresentation.upcastPayload(
+                outputType,
+                org.dom4j.Document::class.java
+            ) { document ->
+                val translatedValue = when (val oldValue = document.rootElement.element("identificationMethod").text) {
+                    "NORWEGIAN_BANK_ID" -> "com.zignsec:BankIDNO"
+                    "DANISH_BANK_ID" -> "com.zignsec:NemID"
+                    else -> oldValue
+                }
```
vonElfvin, line 61: Nice!

member-service/src/main/java/com/hedvig/memberservice/identity/IdentityEventListener.kt
```diff
@@ -1,54 +0,0 @@
...
-import org.axonframework.config.ProcessingGroup
-import org.axonframework.eventhandling.EventHandler
-import org.springframework.data.repository.findByIdOrNull
-import org.springframework.stereotype.Component
-
-@Component
-@ProcessingGroup("IdentifiedMembers")
-class IdentityEventListener(
-    private val repository: IdentityRepository
-) {
```
vonElfvin, line 19: If we don't replace this with something else. Like a `isIdentified` boolean on the `member_entity` then I think we should keep this around for the time being. Otherwise Hope will expect these identities to exist

member-service/src/main/java/com/hedvig/memberservice/query/MemberEventListener.java
vonElfvin: Maybe we can add `isIdentified` here so can safely kill the `MemberIdentity` table.

member-service/src/main/java/com/hedvig/memberservice/events/MemberIdentifiedEvent.kt
vonElfvin: Makes sense, I just explained the thought of the previous developer

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
vonElfvin: Maybe? I wasn't 100% confident in the event's existance though. My thinking was that since a member "is identified" wether we want it or not given that they for example login with bank id then maybe we don't need the event itself. This event does not exist for Swedish members right now for example. Maybe a future thing to address then is to backfil that so we have the event for every member that has ever identified themselves in any way.

Or how do you suggest the event should be considered given past events?

vonElfvin: I think a discussion like this is like 3x easier to have with a whiteboard where you can paint out flows

member-service/src/main/java/com/hedvig/memberservice/services/v2/BankIdServiceV2.kt
vonElfvin: Super nice! How do we handle backfill of this?

member-service/src/test/java/com/hedvig/memberservice/MemberAggregateTests.kt
```diff
@@ -256,7 +262,7 @@ class MemberAggregateTests {
                  MemberIdentifiedEvent(
                     memberId,
                     MemberIdentifiedEvent.NationalIdentification(personalNumber, MemberIdentifiedEvent.Nationality.NORWAY),
-                    MemberIdentifiedEvent.IdentificationMethod.NORWEGIAN_BANK_ID,
+                    "com.zignsec:BankIDNO",
```
vonElfvin, line 265: Do they call it `BankIDNO`? 🤔 😄

member-service/src/test/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdServiceTest.kt
```diff
@@ -5,76 +5,79 @@ import com.hedvig.auth.models.User
...
 import java.util.*
-import org.mockito.Mockito.`when` as whenever
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
 
-@RunWith(MockitoJUnitRunner::class)
 class ZignSecBankIdServiceTest {
 
-    @Mock
+    @MockK
```
vonElfvin, line 27: 💯

member-service/src/test/java/com/hedvig/memberservice/services/signing/zignsec/ZignSecBankIdServiceTest.kt
```diff
@@ -5,76 +5,79 @@ import com.hedvig.auth.models.User
...
-import org.mockito.Mockito.anyLong
-import org.mockito.Mockito.never
-import org.mockito.Mockito.verify
-import org.mockito.junit.MockitoJUnitRunner
-import org.springframework.core.env.Environment
 import java.time.LocalDateTime
 import java.util.*
-import org.mockito.Mockito.`when` as whenever
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
```
vonElfvin, line 23: junit 5 👍

vonElfvin (commented):
This is a really good addition to the login flow! I especially like the upcasting refactorings and a more scaleable model+behaviour overall utilizing strings. 

I like that we added name to Sweden as well, super nice!

One thing though, given the new understanding of the requirements of wether a member "is identified" (logged in with any bankID). We could either:
1. Keep the MemberIdentity projection until we find a replacement for the behaviour
2. Replace the MemberIdentity projection behaviour (maybe with a simple `isIdentified` boolean). Maybe using the actual sign events as well.

Also, how we treat old MemberIdentifications in Sweden? Do we care about them? (They don't have any events today but will now)

PR #491, opened 2021-04-19
[MX-91] Repair user model by deleting login-created users that should've been connected to 'SimpleSigned' members coming from Rapio

bystam:
# Jira Issue: [MX-91]

## What?
- Create a (temporary) importer that handles the `MemberSignedWithoutBankId` events and creates real users for them
- We give these users a `SimpleSignConnection`, since they are technically simple signed (created through Rapio without BankID verification)
- We delete the existing users that was accidentally created upon login and replace them with the new users with the correct connection
- This fix will be tidied up in [MX-92]

## Why?
- Compricer/Insplanet signed members currently can't log in when the head of master is deployed (currently rolled back)

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [ ] Codescouted

[MX-91]: https://hedvig.atlassian.net/browse/MX-91
[MX-92]: https://hedvig.atlassian.net/browse/MX-92


member-service/src/main/java/com/hedvig/auth/models/User.kt
```diff
@@ -20,7 +20,7 @@ class User(
     val id: UUID = UUID.randomUUID()
 
     @OneToMany(mappedBy = "user", cascade = [CascadeType.ALL], orphanRemoval = true)
-    private val auditLog: List<AuditEvent> = listOf()
+    internal val auditLog: List<AuditEvent> = listOf()
```
bystam, line 23: We can make this `private` again once the historical events have been processed and state has been repaired.

member-service/src/main/java/com/hedvig/auth/import/SwedishRepairingSimpleSignMemberImporter.kt
```diff
@@ -0,0 +1,71 @@
...
+    @Autowired private lateinit var auditEventRepository: AuditEventRepository
+    @Autowired private lateinit var userRepository: UserRepository
+    @Autowired private lateinit var simpleSignConnectionRepository: SimpleSignConnectionRepository
+    @Autowired private lateinit var swedishBankIdCredentialRepository: SwedishBankIdCredentialRepository
+
+    private val logger = LoggerFactory.getLogger(javaClass)
+
+    @EventHandler
+    @Transactional
+    fun on(e: MemberSignedWithoutBankId, @Timestamp timestamp: Instant) {
```
bystam, line 31: The goal is to simplify this and merge it with `SwedishMemberImporter` later, but for now we need it to process historical events in its own processing group to make sure it repairs the existing state.

member-service/src/main/java/com/hedvig/auth/import/SwedishRepairingSimpleSignMemberImporter.kt
```diff
@@ -0,0 +1,71 @@
...
+            }
+        }
+
+        val user = User(associatedMemberId = e.memberId.toString(), createdAt = timestamp)
+        user.simpleSignConnection = SimpleSignConnection(
+            user = user,
+            personalNumber = e.ssn,
+            country = "SE",
+            createdAt = timestamp
+        )
```
bystam, line 63: During the bug hunt, we discussed whether it made sense to also consider the "external partner"-signed members to be "simple sign connected". From one perspective, "simple sign" is the name of a feature that was built to enable members to get insurance in the web portal without having to use any kind of BankID.

But from another perspective, "simple signed" could be considered out name for someone who has signed using a personal number but where we have never authenticated it. And with that definition, it makes perfect sense to have said users have a `SimpleSignConnection`, as that is exactly how they are signed.

niho (approved):
This makes sense. Maybe we should rebase this on the refactoring branch? Or at least #492 should be.

---

Some more general thoughts (postmortem, kind of):

- Right now we have multiple different events for when a member signs. Given that, it's not a big surprise that we missed the `MemberSignedWithoutBankId` event (even though I think Elvin mentioned Rapio members at some point).
- Having different event types for different ways of signing makes very little sense to me. It sort of implies that a single member can sign in many different ways. E.g. a timeline where I have both a `DanishMemberSignedEvent` and a `MemberSignedEvent`. What would that even mean? To me it makes much more sense to have a single unified `MemberSignedEvent` that simply looks slightly different depending on the sign method used and has a simple enum property `signMethod`.
- That a member signed is an important core business event that many parts of the Hedvig system might want to listen to in the future. Imagine if we have 20 different markets and 20 different "member signed" events. How would you make sure that you are listening to all of them? How would you keep your event listeners up to date when new markets are added?

niho (approved):

bystam:
> This makes sense. Maybe we should rebase this on the refactoring branch? Or at least #492 should be.

Rebasing it would imply that it has to be merged before the refactoring branch, which I think would be the wrong approach. I think we should aim to get this bugfix out so that we can again have master deployed in production, and then let refactoring happen on top of it.

#492 could absolutely be rebased on it, however.

> Some more general thoughts (postmortem, kind of):
> 
> * Right now we have multiple different events for when a member signs. Given that, it's not a big surprise that we missed the `MemberSignedWithoutBankId` event (even though I think Elvin mentioned Rapio members at some point).
> * Having different event types for different ways of signing makes very little sense to me. It sort of implies that a single member can sign in many different ways. E.g. a timeline where I have both a `DanishMemberSignedEvent` and a `MemberSignedEvent`. What would that even mean? To me it makes much more sense to have a single unified `MemberSignedEvent` that simply looks slightly different depending on the sign method used and has a simple enum property `signMethod`.
> * That a member signed is an important core business event that many parts of the Hedvig system might want to listen to in the future. Imagine if we have 20 different markets and 20 different "member signed" events. How would you make sure that you are listening to all of them? How would you keep your event listeners up to date when new markets are added?

That all makes sense to me. Since it's event sourced it's a little bit more than just refactoring, but AFAIK it's possible throught the means of "upcasting". I guess either we address it in the near future, or we consider it something to revisit upon entering a new market.

vonElfvin:
I agree with you Niklas, I've questioned the fact that we have multiple events before and don't remember what came out of that discussion. We've had bugs in production based solely on the fact that we have multiple events (in addition to this one 😄 ) Might we worth addressing sooner rather than later...

 As Fredrik said, upcasting would hopefully do wonders for a potential refactoring :)

PR #492, opened 2021-04-19
MX-92: Wire the `MemberSignedWithoutBankId` event to the `MemberSimpleSignedEventHandler`

bystam:
# Jira Issue: [MX-92]

## What?
- Merge the temporarily implemented `SwedishRepairingSimpleSignMemberImporter` functionality with the existing `MemberSimpleSignedEventHandler` as they currently produce the same kind of end state.

## Why?
- To keep the code cleaner and avoid keeping complex solutions around.

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [x] Codescouted

[MX-92]: https://hedvig.atlassian.net/browse/MX-92


member-service/src/main/java/com/hedvig/memberservice/events/MemberSimpleSignedEvent.kt
```diff
@@ -1,5 +1,6 @@
 package com.hedvig.memberservice.events
 
+import com.neovisionaries.i18n.CountryCode
```
Michael-Duivestein, line 3: Is this the right import?

bystam: Yes, I got a tip from Elvin that we use that library to get a typesafe way to write country codes. It's quite exhaustive, and I assume very stable.

Michael-Duivestein (approved):

PR #493, opened 2021-04-21
MX-99: repair the users for the oldest Swedish members in our system

bystam:
# Jira Issue: [MX-99]

## What?
- This imports some specific members that were signed pre october 2018 that therefore were missing an SSN value for their `MemberSignedEvent`.

## Why?
- The Hedvig members did not get users to log in to, and therefore now create faulty ones.

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [x] Codescouted

[MX-99]: https://hedvig.atlassian.net/browse/MX-99


member-service/src/main/java/com/hedvig/auth/import/OldLoyalSwedishMemberRepairingImporter.kt
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * This migration targets members that were signed without an SSN value.
+ * This is the case for members who signed before 2018-10-09, where the SSN value on said
+ * event was introduced.
+ *
+ * In order to solve this, we must create them by stitching together some other concepts,
+ * and delete any users that were created upon login because their real ones were missing.
+ */
+@Component
+@ProcessingGroup("OldLoyalSwedishMemberImporter")
```
vonElfvin: Loyal 👑

member-service/src/main/java/com/hedvig/auth/import/OldLoyalSwedishMemberRepairingImporter.kt
```diff
@@ -0,0 +1,80 @@
...
+    @Autowired private lateinit var userRepository: UserRepository
+    @Autowired private lateinit var swedishBankIdCredentialRepository: SwedishBankIdCredentialRepository
+    @Autowired private lateinit var signedMemberRepository: SignedMemberRepository
+
+    private val logger = LoggerFactory.getLogger(javaClass)
+
+    @EventHandler
+    @Transactional
+    fun on(e: MemberSignedEvent, @Timestamp timestamp: Instant) {
+        if (e.ssn != null) return
```
vonElfvin, line 43: Big 🧠 No timestamp or similar shenanigans with the same outcome

member-service/src/main/java/com/hedvig/auth/import/OldLoyalSwedishMemberRepairingImporter.kt
```diff
@@ -0,0 +1,80 @@
...
+
+        if (userRepository.findByAssociatedMemberId(e.id.toString()) != null) {
+            logger.info("Member ${e.id} was already imported - skipping")
+            return
+        }
+
+        val personalNumber = signedMemberRepository.findById(e.id)?.orElse(null)?.ssn ?: return
```
vonElfvin, line 50:
```suggestion
        val personalNumber = signedMemberRepository.findByIdOrNull(e.id)?.ssn ?: return
```

member-service/src/main/java/com/hedvig/memberservice/events/MemberSignedEvent.kt
```diff
@@ -0,0 +1,18 @@
+package com.hedvig.memberservice.events
+
+data class MemberSignedEvent(
+    val id: Long,
+    val referenceId: String,
+    val signature: String,
+    val oscpResponse: String,
```
vonElfvin, line 7: These two fields were added on January 25th 2018, I presume that is fine to assume being not nullable since we signed our first member in February 2018

vonElfvin (commented):
Really good! This will do the trick

member-service/src/main/java/com/hedvig/memberservice/events/MemberSignedEvent.kt
bystam: Oh! Thanks for looking it up. I managed to run it locally on staging data at least. And I assume there's a good chance staging data is about as old as production?

member-service/src/main/java/com/hedvig/auth/import/OldLoyalSwedishMemberRepairingImporter.kt
bystam: Doh! I missed that this was the regular built-in ID, because the `findById` one was declared on the interface itself
image

bystam:
Obviously, once this has been run we should delete it so that we don't have to take it into account in the Auth modularisation PR #489

vonElfvin:
> Obviously, once this has been run we should delete it so that we don't have to take it into account in the Auth modularisation PR #489

We could actually include this login in the regular `MemberSignedImporter` or w/e it's called. To have this fallback for those where ssn is null 🤔

niho (approved):
Super clean! 🏅⭐️

PR #494, opened 2021-04-22
Fix/application context load

niho:
## What?
- Add test to check that the application context loads successfully.
- Delete the `AuthApplication` to make the context load.
- Remove the tracking token for the "com.hedvig.memberservice.users" event processor.

## Why?
- Application context failed to load because of `AuthApplication`.

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [x] Codescouted


bystam (approved):
I love it! Just one suggestion about using fake config for a secret

bystam (approved):

PR #495, opened 2021-04-26
MX-95: GDPR "Right to be forgotten"

niho:
# Jira Issue: [MX-95]

## What?
- [MX-96]: Add a "delete member" internal endpoint
- [MX-98]: Ensure that logs do not contain PII data

### MX-96
- Axon does not support deleting the immutable history of a root aggregate.
- The aggregate is "marked as deleted" (for manual cleanup) and the projections are deleted by listening to `MemberDeletedEvent`.
- See [this Stackoverflow thread](https://stackoverflow.com/questions/63689545/axon-framewok-vs-gdpr-delete-phisically-personal-data-from-domain-event-entry-t?noredirect=1&lq=1) in particular for discussion on the problem.
- An ["Axon GDPR Data Protection"](https://axoniq.io/product-overview/axon-data-protection) module is available as a paid add-on that solves the problem using encryption.

### MX-98
- I have solved the logging of PII data problem by ensuring PII data is only logged:
   - Using the "debug" level, and thus not included by default in production logs (but still available if needed for debug purposes by setting the desired log level to "debug" or "trace").
   - Or using the "error" level, which is fair use according to GDPR (e.g., finding and fixing errors in exceptional cases).
- This model should both be aligned with the spirit of GDPR and the way [log levels are intended to be used](https://sematext.com/blog/logging-levels/).

[MX-95]: https://hedvig.atlassian.net/browse/MX-95
[MX-96]: https://hedvig.atlassian.net/browse/MX-96
[MX-98]: https://hedvig.atlassian.net/browse/MX-98


bisnode-bci/src/main/java/com/hedvig/external/bisnodeBCI/BisnodeClient.java
```diff
@@ -49,7 +49,7 @@ public PersonSearchResultListResponse match(String ssn) throws RestClientExcepti
             entity,
             PersonSearchResultListResponse.class);
 
-    logger.info(response.toString());
+    logger.debug(response.toString());
```
Michael-Duivestein, line 52: Nice!

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
```diff
@@ -5,56 +5,9 @@ import org.axonframework.commandhandling.model.AggregateLifecycle.apply
...
-import com.hedvig.memberservice.commands.UpdateAcceptLanguageCommand
-import com.hedvig.memberservice.commands.UpdateBirthDateCommand
-import com.hedvig.memberservice.commands.UpdateEmailCommand
-import com.hedvig.memberservice.commands.UpdatePhoneNumberCommand
-import com.hedvig.memberservice.commands.UpdatePickedLocaleCommand
-import com.hedvig.memberservice.commands.UpdateSSNCommand
-import com.hedvig.memberservice.commands.UpdateSwedishWebOnBoardingInfoCommand
-import com.hedvig.memberservice.commands.ZignSecSignCommand
-import com.hedvig.memberservice.commands.ZignSecSuccessfulAuthenticationCommand
+import com.hedvig.memberservice.commands.*
```
Michael-Duivestein, line 8: You may want to change the IDE setting to not use * imports:
Settings -> Editor -> Code Style -> Java:
Imports tab:
Change both of these to 999:
`Class count to use import with '*':`
`Names count to use static import with '*':`

niho: Why not use * imports?

Michael-Duivestein: We've been trying to get rid of them everywhere. I can't remember the reasoning off-hand. Will start a thread in #platform.

vonElfvin: Kltint complains about this, that's the only reason I know of

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
```diff
@@ -631,4 +590,9 @@ class MemberAggregate() {
     fun on(e: PickedLocaleUpdatedEvent) {
         member = member.copy(pickedLocale = e.pickedLocale)
     }
+
+    @EventSourcingHandler
+    fun on(e: MemberDeletedEvent) {
+        markDeleted()
```
vonElfvin, line 596: Didn't know this existed! Very smart 🧠

member-service/src/main/java/com/hedvig/memberservice/identity/IdentityEventListener.kt
```diff
@@ -49,6 +50,11 @@ class IdentityEventListener(
             }
         } ?: repository.save(identityEntity)
     }
+
+    @EventHandler
+    fun on(e: MemberDeletedEvent) {
+        repository.deleteById(e.memberId)
+    }
```
vonElfvin, line 57: With `markDeleted  - Marks this aggregate as deleted, instructing a repository to remove that aggregate at an appropriate time.`. Is there a race condition here where the event could be deleted before being handled? 🤔

member-service/src/main/java/com/hedvig/memberservice/query/MemberEventListener.java
```diff
@@ -271,4 +251,18 @@ void on(PickedLocaleUpdatedEvent e) {
     m.setPickedLocale(e.getPickedLocale());
     userRepo.save(m);
   }
+
+  @EventHandler
+  @Transactional
+  void on(MemberDeletedEvent e) {
+      if (userRepo.findById(e.getMemberId()).isPresent()) {
```
vonElfvin, line 258: Can we rename this repository, super confusing name now with the user concept 😄 
```suggestion
      if (memberRepository.findById(e.getMemberId()).isPresent()) {
```

member-service/src/main/java/com/hedvig/memberservice/users/NorwegianMemberSignedEventHandler.kt
```diff
@@ -38,16 +38,16 @@ class NorwegianMemberSignedEventHandler {
         zignSecNotification: ZignSecNotificationRequest
     ) {
         if (userService.findUserByAssociatedMemberId(memberId) != null) {
-            logger.info("Member $memberId already has a user - skipping")
+            logger.debug("Member $memberId already has a user - skipping")
```
vonElfvin, line 41: What are the benefits you see to changing info to debug? I'm thinking it's easier to filter these out in datadog, can you see any more benefits? Should we default to `debug` for loggings like this in your opinion? For me that makes sense actually

member-service/src/test/java/com/hedvig/memberservice/query/MemberEventListenerTest.kt
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Test
+    fun `should delete member entity on MemberDeletedEvent`() {
+        val memberId = 123L
+        eventListener.on(MemberCreatedEvent(memberId, MemberStatus.ONBOARDING))
+        entityManager.flush()
+        entityManager.clear()
+        eventListener.on(MemberDeletedEvent(memberId))
+        assertThat(memberRepository.findById(memberId))
```
vonElfvin, line 51: What does this assert? 🤔

vonElfvin (approved):
Scary stuff! Very clean implementation, added some thoughts to consider but nothing that changes the outcome 🙏

member-service/src/main/java/com/hedvig/memberservice/query/MemberEventListener.java
niho: Sure, I can refactor it. This is very ancient legacy code and not something I've written. I just used the already existing variable.

member-service/src/main/java/com/hedvig/memberservice/users/NorwegianMemberSignedEventHandler.kt
niho: From the article on logging levels that I linked to:

> DEBUG – less granular compared to the TRACE level, but it is more than you will need in everyday use. The DEBUG log level should be used for information that may be needed for diagnosing issues and troubleshooting or when running application in the test environment for the purpose of making sure everything is running correctly

> INFO – the standard log level indicating that something happened, application entered a certain state, etc. For example, a controller of your authorization API may include an INFO log level with information on which user requested authorization if the authorization was successful or not. The information logged using the INFO log level should be purely informative and not looking into them on a regular basis shouldn’t result in missing any important information.

So a production system should not output debug log events (which is usually the default). You can surgically increase the log level to debug a specific issue or run the system locally in development mode (in which you would typically have the log level set to debug by default).

The log events that I have changed to debug in this PR (especially the ones that output sensitive PII data) are good examples of log events that are not really info (i.e., system status) events, but rather low level debug event that we can use to trace a specific problem when needed.

So to summarise:

- **Production environment**: Log level set to INFO
- **Development environment**: Log level set to DEBUG

member-service/src/test/java/com/hedvig/memberservice/query/MemberEventListenerTest.kt
niho: Good catch! It's a typo. I forgot to add the `isEmpty` assertion.

member-service/src/main/java/com/hedvig/memberservice/identity/IdentityEventListener.kt
niho: The call to `markDeleted()` does not delete the events. It only marks the aggregate as deleted. So no, there is no race-condition.

member-service/src/main/java/com/hedvig/memberservice/web/InternalMembersController.java
```diff
@@ -179,5 +174,11 @@ public void editMemberInfo(
         //return ResponseEntity.notFound().build();
 
     }
+
+    @DeleteMapping("/{memberId}")
+    public ResponseEntity<Boolean> deleteMember(@PathVariable Long memberId) {
+        commandBus.sendAndWait(new DeleteMemberCommand(memberId));
+        return ResponseEntity.ok(true);
+    }
```
bystam, line 182: This would simply print the string `"true"` as the response body. Doesn't it make more sense to simply return an empty 200 for success?

```suggestion
    @DeleteMapping("/{memberId}")
    public ResponseEntity<Void> deleteMember(@PathVariable Long memberId) {
        commandBus.sendAndWait(new DeleteMemberCommand(memberId));
        return ResponseEntity.ok().build();
    }
```

bystam (approved):
Simple enough - too bad it's harder to change the event log.

Just had one suggestion about the return value of the delete endpoint.

member-service/src/main/java/com/hedvig/memberservice/web/InternalMembersController.java
niho: Well, that's a subtle but interesting point you're making actually. I have bumped into this "issue" many times before myself when designing REST APIs. The short answer is that returning an empty response body makes the response invalid JSON. So then you need to explicitly set "content-type: text/plain" for this end-point and all clients need to special case it. If you instead continue to use "content-type: application/json" like all the other end-points, then clients that have a strict JSON parser will fail to parse the response and consider the request failed, even if it actually succeeded.

In my mind there are only two sensible things to return in a case like that. Either an empty object `{}`, which would be the JSON equivalent of an empty response body. Or a boolean `true` or `false`, depending on if anything was deleted (the request can still `200` succeed, even if nothing was actually deleted). And out of those two options, I consider the second one to be the most sensible one for a `DELETE` request like this.

But my reasoning might be completely off on this, so please prove me wrong if you want! 🌝

niho:
@mightyost This is your project, so I'm awaiting your review on this before merging. 💂🏻

mightyost (approved):
Endpoint look good too me :). I have not checked the internals but others have.

One reflection: Is it really good to just convert all logs to debug. Feels like the old info logs gives a lot of good tracing when to troubleshoot old requests if needed (even with the PII masked)?

mightyost:
A question: What happens if I send in a member id that has no events to delete. Looks like I will get a HTTP 200 back, could you please confirm that so I do not have to add logic for handling 404 or other return codes on my side...

PR #501, opened 2021-05-05
add an endpoint for resetting the member->user export to a specific point in time

bystam:
# Jira Issue: [MX-125]

## What?
- We need to reset member->user export to a specific point in time

## Usage
```sh
curl -X PUT .../dangerously/reset/MemberToUserExport?fromTime=2020-04-04T10:00:00Z
```

## Why?
- During a buggy deployment, some of them failed to be exported, and now we need to repair it

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [ ] Codescouted

[MX-125]: https://hedvig.atlassian.net/browse/MX-125


vonElfvin (approved):
👍 👍 👍

PR #504, opened 2021-05-12
MX-199: Create properly populated members from a distribtuor point of view

niho:
# Jira Issue: [MX-119]

## What?
- Adds functionality to the Members API to both create a new member through a simple `POST /_/member/` operation
- Adds functionality to populate a member with data through a standard `PATCH /_/member/{id}` operation (reusing existing commands and events)
  - Name
  - SSN
  - Address
  - Contact info (mail/phone)
  - birth date (can be auto-set through SSN)
- Adds proper integration tests for the members API (including things like search)
- Rewrite the members controller to Kotlin (and merges it with the InternalMembersControllerKt that was a single endpoint written in kotlin before)

## Why?
- As part of our AVY partnership, they need to be able to set up members through a public API without doing the regular "a member is auto-created as a side effect of signing an insurance"-dance.
- This will be combined with the Trial insurance API in product pricing (https://github.com/HedvigInsurance/product-pricing/pull/770)

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-119]: https://hedvig.atlassian.net/browse/MX-119


auth/pom.xml
```diff
@@ -56,6 +56,10 @@
             <artifactId>spring-boot-starter-test</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+        </dependency>
```
bystam, line 62: The user module needed this to be able to have endpoints of its own

member-service/src/main/java/com/hedvig/memberservice/commands/UpdateMemberDataCommand.kt
```diff
@@ -1,14 +1,18 @@
...
 import java.time.LocalDate
 
-class MemberUpdateContactInformationCommand(
+class UpdateMemberDataCommand(
     @TargetAggregateIdentifier
     val memberId: Long,
     val firstName: String?,
     val lastName: String?,
+    val ssn: String?,
+    val countryCode: CountryCode?,
```
bystam, line 15: `countryCode` is only here because the `SSNUpdatedEvent` needed a `nationality` value. It's not actually set on the member in any other way.

auth/src/main/kotlin/com/hedvig/auth/models/AuditEvent.kt
```diff
@@ -21,6 +21,7 @@ internal class AuditEvent(
 
     enum class EventType {
         CREATED_ON_IMPORT,
-        CREATED_ON_LOGIN
+        CREATED_ON_LOGIN,
+        CREATED_EXPLICITLY,
```
vonElfvin, line 25: Like the clarity! Maybe even deliberately? 😄 Regardless good naming
```suggestion
        CREATED_EXPLICITLY,
```

member-service/src/main/java/com/hedvig/memberservice/aggregates/LivingAddress.kt
```diff
@@ -1,6 +1,6 @@
 package com.hedvig.memberservice.aggregates
 
-class LivingAddress(
+data class LivingAddress(
```
vonElfvin, line 3: Noice

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
```diff
@@ -225,54 +224,51 @@ class MemberAggregate() {
...
+        if (needsAddressUpdate) {
             apply(
                 LivingAddressUpdatedEvent(
                     id,
                     cmd.street,
                     cmd.city,
                     cmd.zipCode,
                     cmd.apartmentNo,
-                    floor
+                    cmd.floor ?: 0
```
vonElfvin, line 263: Isn't `null` better than saying floor 0?

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
```diff
@@ -225,54 +224,51 @@ class MemberAggregate() {
...
-            apply(BirthDateUpdatedEvent(id, cmd.birthDate))
-        }
     }
 
     @CommandHandler
     fun handle(cmd: SwedishBankIdSignCommand) {
-        if (cmd.personalNumber != null &&
-            member.ssn != cmd.personalNumber
-        ) {
+        if (member.ssn != cmd.personalNumber) {
```
vonElfvin, line 271: Nice code refactoring. But, a thing that struck me, shouldn't the `member.ssn` be regarded as a "purer" source of truth than the `cmd.personalNumber`? This does look a quite risky actually IMO

member-service/src/main/java/com/hedvig/memberservice/web/InternalMembersController.kt
```diff
@@ -0,0 +1,245 @@
...
+
+    @GetMapping("/{memberId}/pickedLocale")
+    fun getPickedLocale(@PathVariable memberId: Long): ResponseEntity<PickedLocaleDTO> {
+        val member = memberRepository.findById(memberId)
+        if (member.isPresent && member.get().pickedLocale != null) {
+            val res = PickedLocaleDTO(member.get().pickedLocale)
+            return ResponseEntity.ok(res)
+        }
+        //TODO: Returning sv_SE per default if member not found for launch in order to not block
+        // notification service.
```
vonElfvin, line 223: This is not good 😢  (I know it's not you who put it here)  

For future reference, we should crash here tbh. I think we have fixed the backfill for this so it can be removed.

member-service/src/main/java/com/hedvig/memberservice/web/v2/MembersControllerV2.kt
```diff
@@ -32,15 +32,8 @@ class MembersControllerV2 @Autowired constructor(
...
-            var memberId: Long?
-            var member: Optional<MemberEntity>
-
-            do {
-                memberId = abs(this.randomGenerator.nextLong() % 1000000000)
-                member = memberRepository.findById(memberId)
-            } while (member.isPresent)
-
-            commandGateway.send<Void>(CreateMemberCommand(memberId!!, acceptLanguage))
+            val memberId = memberIdGenerator.nextId()
```
vonElfvin, line 35: Nice!

vonElfvin (approved):
Great tests as always! 😍 Also really nice to see so much `java` -> `kotlin` 💯 

Added some thoughts that might be worth considering, nothing has to change the outcome of this PR though

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
bystam: Maybe? We didn't change this - it worked like that before. 🤷

member-service/src/main/java/com/hedvig/memberservice/aggregates/MemberAggregate.kt
bystam: Do you mean that SSN only should be allowed to be filled if it was null before?

member-service/src/main/java/com/hedvig/memberservice/web/InternalMembersController.kt
bystam: How do you like this update? :)


image

PR #506, opened 2021-05-19
Fix/gdpr delete me on gdpr deleted members

vonElfvin:
# Jira Issue: []

## What?
- Do not use delete event on members who's name is GDPR (the old way of gdpr deletion)

## Why?
- Since these events are not valid the aggregate cannot be created properly, thus breaking.
image
image

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally
- [ ] Codescouted


mightyost (approved):
Looks good!

PR #509, opened 2021-05-26
Fix name comparison

vonElfvin:
# Jira Issue: []

## What?
- Use `!equals` instead of `!=` for GDPR name comparison

## Why?
- Java

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [x] Codescouted


member-service/src/main/java/com/hedvig/memberservice/web/InternalMembersController.java
```diff
@@ -176,7 +176,12 @@ public void editMemberInfo(
...
-        if (member != null && member.getFirstName() != "GDPR" && member.getLastName() != "GDPR") {
+        if (member != null
+                && member.getFirstName() != null
+                && !member.getFirstName().equals("GDPR")
+                && member.getLastName() != null
+                && !member.getLastName().equals("GDPR")
+        ) {
             commandBus.sendAndWait(new DeleteMemberCommand(memberId));
             return ResponseEntity.ok(true);
         } else {
```
mightyost, line 187: This implies that we can only delete members that has a `firstName` and `lastName` sat. That is maybe always the case but maybe this is more accurate?
```
if (member != null && !"GDPR".equals(member.getFirstName())) {
    commandBus.sendAndWait(new DeleteMemberCommand(memberId));
...
```

mightyost (approved):
Gr8!

And thanks for the learning session looking at this today!!

PR #510, opened 2021-05-26
Fix NPE and return value for unsign endpoint

tobiasbexelius:
# Jira Issue: [MX-146]

## What?
- Fix NPE in unsign member endpoint when member for given ssn does not exist
- Fix return value so that true is only returned when something is deleted

## Why?
- 

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally
- [x] Codescouted

[MX-146]: https://hedvig.atlassian.net/browse/MX-146


member-service/src/main/java/com/hedvig/memberservice/services/QualityAssuranceService.kt
```diff
@@ -25,7 +25,7 @@ class QualityAssuranceService(
 
     private fun deleteSignedMemberEntity(ssn: String): Boolean {
         val signedMemberEntity = signedMemberRepository.findBySsn(ssn)
-        if (signedMemberEntity != null) {
+        if (signedMemberEntity.isPresent) {
```
vonElfvin, line 28: Nice catch!!

🕵️‍♂️

member-service/src/main/java/com/hedvig/memberservice/services/QualityAssuranceService.kt
```diff
@@ -44,8 +44,11 @@ class QualityAssuranceService(
     private fun deleteUser(ssn: String): Boolean {
         val member = memberRepository.findBySsn(ssn).firstOrNull()
```
vonElfvin, line 45: I think it's likely that this will find the wrong member. Since it might find a member with the same ssn that was not the one signed. Maybe @bystam can shed some light on that though?

vonElfvin: Or we can delete all users for all members instead of the first one. Unsure what the risk is in that though, probably worth 🤷

vonElfvin (commented):
Cool! I could see a potential bug in the code we might want to look into 🙏

member-service/src/main/java/com/hedvig/memberservice/services/QualityAssuranceService.kt
bystam: I guess we could use the `DeleteMember` command here, yes. That could probably even replace the "delete user" one because it will delete the user implicitly (but I guess not very obviously).

I don't think there's one super obvious answer here... the fact that we have many members with the same SSN makes several things non-obvious in this regard. Especially since the UI for "unsign" takes a personal number and not a member ID.

I think actually deleting members could cause "system state integrity" issues though. Other services won't know that the member is deleted, and they might rely upon it still.

I think Elvins proposal that we fetch **all** the members for an SSN value and delete all the users we find for those members is the most robust solution.

bystam (changes requested):
I propose doing the change as suggested by Elvin: find **all** members based on SSN and delete all the users you find connected to it (should only be one, but that way we make sure we find it).

tobiasbexelius:
Fixed!

bystam (approved):
Awesome!

PR #511, opened 2021-05-26
Fix/restore gdpr wrongly deleted member

mightyost:
## What?
Added a new temporary endpoint for reviving accidentally deleted members.

The events for member to revive is fetched and replayed.

## Optional checklist
- [x] Unit tests written
- [x] Tested locally
- [ ] Codescouted


member-service/src/main/java/com/hedvig/memberservice/services/ReviveMemberService.java
```diff
@@ -0,0 +1,204 @@
...
+            throw new IllegalStateException("Member has no contract");
+        }
+
+        List<? extends DomainEventMessage<?>> events = getMemberEvents(memberId);
+
+        if (!isDeleted(events)) {
+            throw new IllegalStateException("Member is not deleted");
+        }
+
+        log.info("Replaying {} events for member", events.size());
```
bystam, line 82: I know the event sequence will make it discernible, but maybe we should either put the memberId in the MDC here or include them in all log statements? Otherwise, a log message only containing `Replaying 17 events for member` won't really tell us much on its own. What do do you think?

member-service/src/main/java/com/hedvig/memberservice/services/ReviveMemberService.java
```diff
@@ -0,0 +1,204 @@
...
+        return memberRepository.findById(memberId).orElse(null);
+    }
+
+
+    private List<? extends DomainEventMessage<?>> getMemberEvents(Long memberId) {
+        return eventStore.readEvents(memberId.toString()).asStream().collect(Collectors.toList());
+    }
+
+    private boolean isDeleted(List<? extends DomainEventMessage<?>> events) {
+        return events.stream().anyMatch(e -> e.getPayloadType().getSimpleName().equals("MemberDeletedEvent"));
```
bystam, line 201: Do we intend this to be expensive? I would assume `MemberDeletedEvent` to reside at the back of the events, and them maybe searching the stream in a reversed manner would make sense?

member-service/src/test/java/com/hedvig/memberservice/users/ReviveDeletedMemberTest.kt
```diff
@@ -0,0 +1,403 @@
...
+            )
+        )
+
+        assertThrows<IllegalStateException> {
+            reviveMemberService.revive(123)
+        }
+    }
+
+    @Test
+    fun `test member is deleted`() {
```
vonElfvin, line 364: Great test coverage! I'd love to see tests for `UserEntity` as well with the right credentials. Otherwise, the member won't be able to login to the App :(

vonElfvin (commented):
Super duper nice!

The only two things I'd consider:
1. Adding a transaction in case the handling of some event would fail, thus breaking the chain
2. Adding tests/validations for `UserEntity` so we know the member can login to the App

member-service/src/test/java/com/hedvig/memberservice/MemberRegistryApplicationTest.kt
```diff
@@ -1,11 +1,17 @@
...
+import com.hedvig.memberservice.services.customerio.CustomerIOEventListener
+import com.ninjasquad.springmockk.MockkBean
 import org.junit.jupiter.api.Test
 import org.springframework.boot.test.context.SpringBootTest
 
 @SpringBootTest
 class MemberRegistryApplicationTest {
 
+    @MockkBean(relaxed = true)
+    private lateinit var reviveMemberService: ReviveMemberService
```
bystam, line 13: Does this one need to be mocked? The entire purpose of `MemberRegistryApplicationTest` is that it's meant to boot the service for real, to make sure that it works. Adding MockBeans to it kind of defeats the purpose of the test.

bystam (commented):
This looks like a proper "driving a motocycle without a helmet" kind of PR - but great nontheless :D 

I had a few quick thoughts.

member-service/src/test/java/com/hedvig/memberservice/MemberRegistryApplicationTest.kt
bystam: The test kind of exists as a protection against adding code that requires new configuration but not checking in said configuration into the application.yml file.

member-service/src/main/java/com/hedvig/memberservice/services/ReviveMemberService.java
```diff
@@ -0,0 +1,204 @@
...
+                    break;
+
+                case "LivingAddressUpdatedEvent":
+                    memberEventListener.on((LivingAddressUpdatedEvent) event.getPayload());
+                    break;
+
+                case "NameUpdatedEvent":
+                    memberEventListener.on((NameUpdatedEvent) event.getPayload());
+                    customerIOEventListener.on((NameUpdatedEvent) event.getPayload());
+                    break;
```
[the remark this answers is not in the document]
meletios: Bot-service is not in our scope for now.

member-service/src/test/java/com/hedvig/memberservice/users/ReviveDeletedMemberTest.kt
mightyost: Would love to test it but I failed to autowire the service bean(s) needed. Will consult @bystam later, maybe we can stitch it together then...

member-service/src/main/java/com/hedvig/memberservice/services/ReviveMemberService.java
```diff
@@ -0,0 +1,204 @@
...
+import com.hedvig.memberservice.users.MemberSignedEventHandler;
+import com.hedvig.memberservice.users.MemberSimpleSignedEventHandler;
+import com.hedvig.memberservice.users.NorwegianMemberSignedEventHandler;
+import org.axonframework.eventsourcing.DomainEventMessage;
+import org.axonframework.eventsourcing.eventstore.EventStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Service;
+
+import java.util.*;
```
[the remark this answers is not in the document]
mightyost: Got this even if my IntelliJ is configured to not behave like this :(.

There where a discussion on Slack about this that never was concluded. I am in the camp to use the "*" but that is a different story ;)...

member-service/src/main/java/com/hedvig/memberservice/services/ReviveMemberService.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    public MemberEntity revive(Long memberId) {
+
+        log.info("Revive deleted member {}", memberId);
+
+        MemberEntity member = memberRepository.findById(memberId).orElse(null);
+        if (member != null) {
+            throw new IllegalStateException("Member already exists");
+        }
```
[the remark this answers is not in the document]
mightyost: Changed to use `exists`, did not know that trick.

I have not added the id to exception, I will rely on having it in the MDC instead...

member-service/src/main/java/com/hedvig/memberservice/services/ReviveMemberService.java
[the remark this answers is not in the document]
mightyost: No, the GDPR job do not delete members in bot service.

member-service/src/test/java/com/hedvig/memberservice/users/ReviveDeletedMemberTest.kt
```diff
@@ -0,0 +1,403 @@
...
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest
+import org.springframework.test.context.junit4.SpringRunner
+import java.lang.IllegalStateException
+import java.time.Instant
+import java.time.LocalDate
+import java.util.*
+
+@DataJpaTest
+@RunWith(SpringRunner::class)
```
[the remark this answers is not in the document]
mightyost: Nice, thanks for the pointers, works like a charm.

member-service/src/test/java/com/hedvig/memberservice/MemberRegistryApplicationTest.kt
mightyost: Thanks a lot, it failed without it being mocked which was a warnings sign that I ignored :). It failed due to a `ReviveMemberService` dependency on `CustomerIOEventListener` that is only enabled by running with the "customer.io" profile. I managed to make the dependancy optional and now it works...

Maybe do this test for every profile (and not just the default) to catch any startup errors?

member-service/src/main/java/com/hedvig/memberservice/services/ReviveMemberService.java
```diff
@@ -0,0 +1,204 @@
...
+
+        List<? extends DomainEventMessage<?>> events = getMemberEvents(memberId);
+
+        if (!isDeleted(events)) {
+            throw new IllegalStateException("Member is not deleted");
+        }
+
+        log.info("Replaying {} events for member", events.size());
+
+        for (DomainEventMessage<?> event : events) {
```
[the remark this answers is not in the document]
mightyost: @Transactional test written with many thanks to @bystam!

member-service/src/test/java/com/hedvig/memberservice/users/ReviveDeletedMemberTest.kt
mightyost: Fixed now :)!

member-service/src/test/java/com/hedvig/memberservice/users/ReviveDeletedMemberTest.kt
```diff
@@ -0,0 +1,418 @@
...
+//    private lateinit var danishMemberSignedEventHandler: DanishMemberSignedEventHandler
+//
+//    @MockkBean(relaxed = true)
+//    private lateinit var memberSignedEventHandler: MemberSignedEventHandler
+//
+//    @MockkBean(relaxed = true)
+//    private lateinit var memberSimpleSignedEventHandler: MemberSimpleSignedEventHandler
+//
+//    @MockkBean(relaxed = true)
+//    private lateinit var norwegianMemberSignedEventHandler: NorwegianMemberSignedEventHandler
```
vonElfvin, line 85: May as well remove these ones now?

member-service/src/test/java/com/hedvig/memberservice/users/ReviveDeletedMemberTest.kt
```diff
@@ -0,0 +1,418 @@
...
+                    NameUpdatedEvent(memberId, "first", "second"),
+                    MemberDeletedEvent(memberId)
+                )
+            )
+        }
+
+        verify(exactly = 1) { customerIOEventListener.on(any() as EmailUpdatedEvent) }
+        verify(exactly = 1) { customerIOEventListener.on(any() as NameUpdatedEvent) }
+
+        val exists = memberRepository.findById(memberId).isPresent
```
vonElfvin, line 372: 😉 
```suggestion
        val exists = memberRepository.existsById(memberId)
```

mightyost: Thx!

member-service/src/test/java/com/hedvig/memberservice/users/ReviveDeletedMemberTest.kt
mightyost: Thx!

PR #512, opened 2021-05-28
Insert ssn from member if none on MemberSignedEvent

mightyost:


member-service/src/main/java/com/hedvig/memberservice/services/ReviveMemberService.java
```diff
@@ -134,8 +134,26 @@ public MemberEntity revive(Long memberId) {
...
 
                 case "MemberSignedEvent":
-                    memberEventListener.on((MemberSignedEvent) event.getPayload(), event.getTimestamp());
-                    memberSignedEventHandler.on((MemberSignedEvent) event.getPayload());
+                    MemberSignedEvent signedEvent = (MemberSignedEvent) event.getPayload();
+
+                    if (signedEvent.getSsn() == null) {
+                        MemberEntity member = memberRepository.findById(memberId).orElse(null);
+                        if (member != null) {
+
```
vonElfvin, line 142: I'd say a better way is to manually pick out the `SsnUpdatedEvent` and get the ssn from there.

vonElfvin (commented):
Nice bug catch!

meletios (approved):

PR #513, opened 2021-05-31
Added tests for taking ssn from UpdateSSNEvent instead of projection.

mightyost:
- Added tests for taking ssn from UpdateSSNEvent instead of projection when reviving deleted members.

- Migrated the `ReviveMemberService` to Kotlin.

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [x] Codescouted


member-service/src/main/java/com/hedvig/memberservice/services/ReviveMemberService.kt
```diff
@@ -0,0 +1,161 @@
...
+                    var signEvent = event.payload as MemberSignedEvent
+
+                    if (signEvent.ssn == null) {
+
+                        val ssn = events.reversed().stream()
+                            .filter { it.payloadType.simpleName == "SSNUpdatedEvent" }
+                            .map { it.payload as SSNUpdatedEvent }
+                            .map { it.ssn }
+                            .filter { it != null }
+                            .findFirst()
```
vonElfvin, line 114: I think there exists something like
```suggestion
                            .firstNotNull()
```

vonElfvin (approved):
Neat

member-service/src/main/java/com/hedvig/memberservice/services/ReviveMemberService.kt
```diff
@@ -0,0 +1,161 @@
...
+    private val memberRepository: MemberRepository,
+    private val contractsService: ContractsService,
+    private val eventStore: EventStore,
+    private val memberEventListener: MemberEventListener,
+    private val identityEventListener: IdentityEventListener,
+    private val danishMemberSignedEventHandler: DanishMemberSignedEventHandler,
+    private val memberSignedEventHandler: MemberSignedEventHandler,
+    private val memberSimpleSignedEventHandler: MemberSimpleSignedEventHandler,
+    private val norwegianMemberSignedEventHandler: NorwegianMemberSignedEventHandler,
+    customerIOEventListener: Optional<CustomerIOEventListener?>
```
bystam, line 55: Should the inner `CustomerIOEventListener` really be nullable in here? Also, have you checked whether simply doing `private val customerIOEventListener: CustomerIOEventListener?` works here? I think maybe spring+kotlin might have some things working for that.

member-service/src/main/java/com/hedvig/memberservice/services/ReviveMemberService.kt
```diff
@@ -0,0 +1,161 @@
...
+                "PickedLocaleUpdatedEvent" -> memberEventListener.on(event.payload as PickedLocaleUpdatedEvent)
+                "NewCashbackSelectedEvent" -> memberEventListener.on(event.payload as NewCashbackSelectedEvent)
+                "MemberIdentifiedEvent" -> identityEventListener.on((event.payload as MemberIdentifiedEvent))
+                "EmailUpdatedEvent" -> {
+                    memberEventListener.on(event.payload as EmailUpdatedEvent)
+                    customerIOEventListener?.on((event.payload as EmailUpdatedEvent))
+                }
+                "NameUpdatedEvent" -> {
+                    memberEventListener.on(event.payload as NameUpdatedEvent)
+                    customerIOEventListener?.on((event.payload as NameUpdatedEvent))
```
bystam, line 102: Some methods in here have an extra parens around its input. Is there a specific reason for that?

member-service/src/test/java/com/hedvig/memberservice/users/ReviveDeletedMemberTest.kt
```diff
@@ -372,6 +372,30 @@ class ReviveDeletedMemberTest {
...
+
+        assertThat(member.id).isEqualTo(memberId)
+        assertThat(member.status).isEqualTo(MemberStatus.SIGNED)
+
+        val signedMember = signedMemberRepository.findById(memberId).get()
+        assertThat(signedMember.ssn).isEqualTo(ssn)
+
+        val user = userService.findUserByAssociatedMemberId(memberId.toString())
+        assertThat(user).isNotNull()
+    }
```
bystam, line 398: I noticed (further up than this diff) that this is a `@SpringBootTest` but without either `@Transactional` or manual database cleanups. This means that any changes made in these tests will remain in the DB for other tests, which probably isn't very good.

member-service/src/test/java/com/hedvig/memberservice/users/ReviveDeletedMemberTest.kt
```diff
@@ -372,6 +372,30 @@ class ReviveDeletedMemberTest {
...
+
+        val memberId = Random().nextLong()
+        val ssn = Random().nextLong().toString()
+
+        val member = execute(memberId,
+            listOf(
+                MemberCreatedEvent(memberId, MemberStatus.INITIATED),
+                SSNUpdatedEvent(memberId, ssn, SSNUpdatedEvent.Nationality.NORWAY),
+                MemberSignedEvent(memberId, "refid", "signature", "oscp", null),
+                MemberDeletedEvent(memberId)
```
bystam, line 386: Should test really have a `MemberDeletedEvent`?

bystam (commented):
Looks good! Added some thoughts

member-service/src/test/java/com/hedvig/memberservice/users/ReviveDeletedMemberTest.kt
bystam: Marking the class with `@Transactional` means spring will run everything in a transaction and roll it back. But you could also manually just delete things from the repository :)

member-service/src/main/java/com/hedvig/memberservice/services/ReviveMemberService.kt
mightyost: Had to make the `customerIOEventListener` an `Optional` since it is only available in the profile "customer.io" used in prod. Starting without that profile would fail the server to start...

Did not understand your second suggestion how to set the `customerIOEventListener`?

member-service/src/main/java/com/hedvig/memberservice/services/ReviveMemberService.kt
mightyost: Thanks, removed those.

member-service/src/main/java/com/hedvig/memberservice/services/ReviveMemberService.kt
mightyost: Yeah, I have seen a `.filterNotNull()` somewhere else but I cannot find anything like it on the stream using code completion...

member-service/src/test/java/com/hedvig/memberservice/users/ReviveDeletedMemberTest.kt
mightyost: Yes. It will not be possible to revive a member not having that event...

member-service/src/test/java/com/hedvig/memberservice/users/ReviveDeletedMemberTest.kt
mightyost: Yeah, good info. I have created the tests to use random ids and not reusing any fixed ones that would make test to conflict with others. Tried with `@Transactional` since thats a better pattern, however my test testing the `@Transactional` on the `ReviveMemberService` fails to run then...

PR #515, opened 2021-06-14
Disable the possibility of changing the SSN value on a member once it is already set

bystam:
# Jira Issue: []

## What?
- Take `OnboardingStartedWithSSNEvent`, `NorwegianSSNUpdatedEvent` and `DanishSSNUpdatedEvent`, all of which are basically some form of "ssn updated" event, and collapse them into the single existing `SSNUpdatedEvent`, to simplify our event history.
- Add protections to the `MemberAggregate` so that the SSN value of a member can't change after it has been set once.

## Why?
- The fact that SSN is basically mutable on our members is broken. If it were to actually happen, it would set the member in an illegal state.

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [x] Codescouted


member-service/src/main/java/com/hedvig/memberservice/events/SSNUpdatedEvent.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    enum class Nationality {
+        SWEDEN, NORWAY, DENMARK
+    }
+
+    /**
+     * There were basically 3 extra events that were SSNUpdatedEvent and nothing else. This class collapses
+     * all 3 into simply SSNUpdatedEvent.
+     */
+    class CollapsingUpcaster : SingleEventUpcaster() {
```
Michael-Duivestein, line 24:
```suggestion
    class SSNUpdatedCollapsingUpcaster : SingleEventUpcaster() {
```

Michael-Duivestein (commented):
Would it be correct to assume that the SSN is verified in some way before changing is blocked?
In other words, can someone change their SSN if they made a mistake when entering it?
(Do we even want that functionality when SSN is tied to quote prices in some countries?)

member-service/src/main/java/com/hedvig/memberservice/events/SSNUpdatedEvent.kt
bystam: Is that name needed when it's already part of a nested class called `SSNUpdatedEvent` you think?

bystam:
> Would it be correct to assume that the SSN is verified in some way before changing is blocked?

It is all verified in Sweden, but not for "simple signed" members in norway and denmark. That's a pretty big gaping hole in our product. It's not a bug so much as it's us creating real live members for data that's is entirely unverified. This PR does nothing to either improve or worsen that.

> In other words, can someone change their SSN if they made a mistake when entering it?

In theory they could before, and this PR would remove that possibility. But if they did then things would stop working randomly in places in our codebase.

> (Do we even want that functionality when SSN is tied to quote prices in some countries?)

The best thing is if the SSN value in `member-service` is decoupled from quotes/pricing and don't have to depend on each other. Optimally, the other services shouldn't need to care about the state of the members when calculating prices.

Regardless, I think they shouldn't be able to change no. Not in the way things look right now. I think the way our members are modeled makes them have very weak data integrity. Not only is SSN mutable, but it's also not unique, so we have no construct that says "this row in the database is this human being". And members, who are the closest thing we have to said thing, can be created without proving their identity. It's quite hard to reason about.

member-service/src/main/java/com/hedvig/memberservice/events/SSNUpdatedEvent.kt
Michael-Duivestein: Can the class be referenced from outside of SSNUpdatedEvent?
If so, the name feels like it could be confusing.

bystam: I mean yes, but it's referenced as `SSNUpdatedEvent.CollapsingUpcaster`

PR #517, opened 2021-06-16
MX-181: include Users in the 'is this SSN taken?' check, not only signed members

bystam:
# Jira Issue: [MX-181]

## What?
- Add cases to the "is this SSN taken?" check that checks whether it is taken by a User, not necessarily a signed member
- Refactor `UserService` to be a bit more compact
- Add some tests for the `PersonController` API
- Add some integration tests for the `SignController` API

## Why?
- The `/v2/member/sign/signedSSN` and `/_/person/has/signed` endpoints are designed to return true for "signed members", a concept that is a little bit vague and old, since we now don't really talk about Members being signed, but contracts.
- People who get created externally from Avy will be Members, and will have Trial insurance, but they will **not have status Signed**. However we still want their personal numbers to be "taken". Therefore we need to find a way to do it. There are basically only bad options here.
- This is awkward because we have a system-wide uniqueness requirement on personal numbers, without any actual system/service/entity enforcing it very clearly.

## Optional checklist
- [x] Unit tests written
- [ ] Tested locally
- [x] Codescouted

[MX-181]: https://hedvig.atlassian.net/browse/MX-181


auth/src/main/kotlin/com/hedvig/auth/services/UserService.kt
```diff
@@ -274,4 +178,35 @@ class UserService {
...
+                user = user,
+                eventType = when (accessType) {
+                    AccessType.LOGIN -> AuditEvent.EventType.CREATED_ON_LOGIN
+                    AccessType.SIGN -> AuditEvent.EventType.CREATED_ON_IMPORT
+                    AccessType.CREATE -> AuditEvent.EventType.CREATED_EXPLICITLY
+                }
+            )
+        )
+        return user
+    }
```
bystam, line 211: Just a little refactor :) made things a bit more compact and clear imo.

member-service/src/main/java/com/hedvig/memberservice/services/signing/SigningService.kt
```diff
@@ -88,10 +88,10 @@ class SigningService(
             throw MemberHasExistingInsuranceException()
         }
         return try {
-            val signMember = commandGateway.send<Any>(
+            commandGateway.sendAndWait<Any>(
                 SignMemberFromUnderwriterCommand(memberId, request.ssn)
             )
-            MemberSignUnderwriterQuoteResponse(memberId, signMember.isDone)
+            MemberSignUnderwriterQuoteResponse(memberId, true)
```
bystam, line 94: I think this implementation was actually not correct. The `signMember.isDone` I think would always return false, since it was happening asynchronously and was a Future.

I changed to `sendAndWait` because it was simpler to test it that way too.

member-service/src/main/java/com/hedvig/personservice/persons/PersonService.kt
```diff
@@ -83,7 +85,30 @@ class PersonService @Autowired constructor(
...
+        // don't accidentally create duplicate "humans" in our system.
+        val credentials = listOf(
+            UserService.Credential.SwedishBankID(personalNumber),
+            UserService.Credential.SimpleSign(countryCode = "SE", personalNumber = personalNumber),
+            UserService.Credential.SimpleSign(countryCode = "NO", personalNumber = personalNumber),
+            UserService.Credential.SimpleSign(countryCode = "DK", personalNumber = personalNumber)
+        )
+        return credentials.any {
+            userService.findUserWithCredential(it) != null
+        }
```
bystam, line 111: SimpleSign is always going to be weird I think :/

member-service/src/test/java/com/hedvig/personservice/persons/web/PersonControllerTest.kt
```diff
@@ -0,0 +1,115 @@
...
+        client.post(
+            uri = "/_/member/$memberId/updateSSN",
+            body = mapOf("ssn" to "201212121212", "nationality" to "SWEDEN")
+        ).assert2xx()
+        // mark it as signed with SSN
+        client.post(
+            uri = "/v2/member/sign/underwriter",
+            body = mapOf("ssn" to "201212121212"),
+            headers = mapOf("Hedvig.token" to memberId.toString())
+        ).assert2xx()
```
bystam, line 67: This member journey kind of resembles the one used when signing members from Rapio.

vonElfvin (approved):
Very good!

member-service/src/test/java/com/hedvig/memberservice/MemberRegistryApplicationTest.kt
```diff
@@ -1,13 +0,0 @@
...
-import org.springframework.boot.test.context.SpringBootTest
-
-@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
-class MemberRegistryApplicationTest {
-
-    @Test
-    fun contextLoads() {
-    }
-
-}
```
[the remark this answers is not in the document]
bystam: We already do that through other integration tests - this one predates them. And it had a bit different context so it took quite a lot of time to run.

